PR #1, opened 2017-06-10
James dev

Mackey22:
Framework in place to automatically setup parsing protocol for any REST API defined in a swagger json file. 

- Uses argparse to create parsers with help display for every endpoint described in an API. 
- Run ```python define_api.py -h``` for a list of top-level commands. 
- Run ```python define_api.py <endpoint> -h``` for a list of command line arguments and their purposes.


Mackey22:
Ready for a code review when you get the chance @kislyuk

samanthachui:
This resolves this issue: https://github.com/HumanCellAtlas/data-store/issues/58

hca/define_api.py
```diff
@@ -0,0 +1,40 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+# import os, sys, json, time, base64
```
kislyuk: Please remove unused commented code.

hca/define_api.py
```diff
@@ -0,0 +1,40 @@
...
+from io import open
+# import pprint
+import requests
+from .parser import get_parser
+
+
+class Api:
+    """Class for interacting with the REST Api."""
+
+    def __init__(self, api_url, user, password):
```
kislyuk: There will not be passwords involved, so let's drop that unused argument.

hca/parser.py
```diff
@@ -0,0 +1,141 @@
...
+        else:
+            param_name = param['name']
+            indexed_parameters[param_name] = param
+    # pprint.pprint(indexed_parameters)
+    return indexed_parameters
+
+
+def get_or_else(d, key, els):
+    """Get key from d. If key not in d, get els."""
+    return d[key] if key in d else els
```
kislyuk: I suggest dropping this function. It's a one-liner that is better written as `d.get(key, els)`, and it has a docstring that adds no information.

setup.py
```diff
@@ -13,6 +13,7 @@
     description='Human Cell Atlas Data Storage System Command Line Interface',
     long_description=open('README.rst').read(),
     install_requires=[
+        "requests==2.17.3"
         'boto3 >= 1.4.4, < 2'
```
kislyuk: This is a syntax error, isn't it? Also, please remove boto3 if we're not going to use it.

test/test.py
```diff
@@ -10,9 +10,155 @@
...
+                "description": "A RFC4122-compliant ID.",
+                "in": "body",
+                "name": "extras-bundle_uuid",
+                "required": True,
+                "type": "string",
+            }}
+        )
+
+    def test_parsing(self):
+        """Test that the parser parses arguments in the right way."""
```
kislyuk: s/in the right way/correctly/

hca/define_api.py
```diff
@@ -0,0 +1,40 @@
...
+# import os, sys, json, time, base64
+import json, sys, os
+from io import open
+# import pprint
+import requests
+from .parser import get_parser
+
+
+class Api:
+    """Class for interacting with the REST Api."""
```
kislyuk: API should probably be all caps throughout. Also, "class API" is not really descriptive. I think what this class does is build a parser, right? How about calling it ParserBuilder? And the docstring could then say something like "Given a Swagger API definition, builds an argparse parser for a CLI".

Mackey22: The parser building all takes place in parser.py. I was intending this class to actually make the requests given the cmd line input.

Makefile
```diff
@@ -1,8 +1,13 @@
 SHELL=/bin/bash
 
+constants: hca/api_spec.json
+
 lint:
 	./setup.py flake8
 
+hca/api_spec.json:
+	wget https://hca-dss.czi.technology/v1/swagger.json
```
kislyuk: Suggest rewriting this to `curl https://hca-dss.czi.technology/v1/swagger.json > hca/api_spec.json`

kislyuk:
LGTM pending the one change that I commented on.

PR #2, opened 2017-06-20
Requests fully implemented

Mackey22:
We should probably go over this together. Grab me when you have time tomorrow.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "checksum_reader"]
 	path = checksum_reader
-	url = git@github.com:HumanCellAtlas/checksum_reader.git
+	url = https://github.com/HumanCellAtlas/checksum_reader.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "checksum_reader"]
 	path = checksum_reader
-	url = https://github.com/HumanCellAtlas/checksum_reader.git
+	url = https://github.com/HumanCellAtlas/checksumming_io.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "checksum_reader"]
+	path = checksum_reader
+	url = git@github.com:HumanCellAtlas/checksum_reader.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "checksum_reader"]
-	path = checksum_reader
+	path = checksumming_io
 	url = https://github.com/HumanCellAtlas/checksumming_io.git
--- a/hca/define_api.py
+++ b/hca/define_api.py
@@ -17,7 +17,8 @@ class API:
     def __init__(self, test=False):
         """Initialize the CLI API."""
         spec = self.get_spec(test)
-        self.base_url = "https://" + spec['host'] + spec['basePath']
+        scheme = spec["schemes"][0] if "schemes" in spec else "https"
+        self.base_url = "{}://{}{}".format(scheme, spec['host'], spec['basePath'])
         self.parser, self.param_data = get_parser(spec)
 
     def get_spec(self, test=False):
--- a/hca/define_api.py
+++ b/hca/define_api.py
@@ -1,6 +1,9 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import json, sys, o
+# import os, sys, json, time, base64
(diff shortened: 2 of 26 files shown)


hca/define_api.py
```diff
@@ -2,36 +2,133 @@
...
 
-    def get_spec(self):
+        self.positional_args = {endpoint: [arg['argument'] for arg in self.param_data[endpoint]['positional']] for endpoint in self.param_data}
+
+    def get_spec(self, test):
         """Load the API specification."""
-        with open(os.path.dirname(os.path.realpath(__file__)) + "/api_spec.json") as fp:
+        url = os.path.dirname(os.path.realpath(__file__)) + "/api_spec.json"
+        if test:
+            url = os.path.dirname(os.path.realpath(__file__)) + "/../test/test.json"
```
Mackey22: Realized I'd never responded to these threads, but I've done each of the changes you recommended. In this case, api_spec is the one downloaded from the remote api whereas test.json is a spec that I've been using to test different features myself.

hca/define_api.py
```diff
@@ -2,36 +2,133 @@
...
         """Initialize the CLI API."""
-        self.spec = self.get_spec()
-        self.base_url = self.spec['host'] + self.spec['basePath']
-        self.parser = get_parser(self.spec)
+        spec = self.get_spec(test)
+        self.base_url = "https://" + spec['host'] + spec['basePath']
+        self.parser, self.param_data = get_parser(spec)
 
-    def get_spec(self):
+        self.positional_args = {endpoint: [arg['argument'] for arg in self.param_data[endpoint]['positional']] for endpoint in self.param_data}
```
Mackey22: 👍

hca/define_api.py
```diff
@@ -2,36 +2,133 @@
...
+import pprint
 import requests
 from .parser import get_parser
 
 
 class API:
     """Class for interacting with the REST Api."""
 
-    def __init__(self, api_url, user):
+    def __init__(self, user, password, test=False):
```
Mackey22: Yeah that makes much more sense. Wasn't sure how we were doing authentication at the beginning - I think that user, pw combo was from back then. Not sure why it got re-added.

hca/full_upload.py
```diff
@@ -0,0 +1,159 @@
...
+
+import sys, os, json
+from io import open
+import argparse
+from datetime import datetime
+import uuid
+import pprint
+
+from .constants import Constants
+from .upload_to_s3 import upload_to_s3
```
ttung: what is your plan for making this multi-cloud?

ttung:
Please consider splitting into separate diffs.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "checksum_reader"]
```
Mackey22: Not in pypi at this point. Would be great to get that up there actually this whole submodule thing is annoying. @sampierson I can look at that with you tomorrow if you have time.

hca/full_upload.py
Mackey22: I'll rename file and class to `upload_to_cloud`.

hca/define_api.py
```diff
@@ -1,37 +1,165 @@
...
+    def __init__(self, test=False):
         """Initialize the CLI API."""
-        self.spec = self.get_spec()
-        self.base_url = self.spec['host'] + self.spec['basePath']
-        self.parser = get_parser(self.spec)
+        spec = self.get_spec(test)
+        self.base_url = "https://" + spec['host'] + spec['basePath']
+        self.parser, self.param_data = get_parser(spec)
+
+    def get_spec(self, test):
```
Mackey22: Will default to False

hca/define_api.py
```diff
@@ -1,37 +1,165 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import json, sys, os
+import json, sys, os, ast
```
Mackey22: Will do. Should I not do one import per line if coming from same module? Also seen standards where w/in package imports are below other imports. Is this something we do as well?

Mackey22:
@ttung will do. Focusing on download for demo tomorrow first then will try to clean this up.

hca/define_api.py
ttung: Yeah, if it's from the same module, you can do multiline.

Mackey22:
Refactored into smaller pull requests.

PR #4, opened 2017-06-30
Added high level commands for the hca console - upload and download. …

Mackey22:
…Both have a run method that is called when that is the argument given and an add_parser method that adds the argument to the parser.

### DIFF
--- a/hca/full_download.py
+++ b/hca/full_download.py
@@ -98,7 +98,7 @@ class FullDownload:
             files = response.json()["bundle"]["files"]
             folder = bundle_name
             if not os.path.isdir(folder):
-                os.mkdir(folder)
+                os.makedirs(folder)
                 logging.info("Bundle {}: GET SUCCEEDED. {} files to download".format(bundle_uuid, len(files)))
             logging.info("Request response:")
             logging.info("{}\n".format(response.content.decode()))
--- a/hca/full_download.py
+++ b/hca/full_download.py
@@ -2,13 +2,9 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import os
 import sys
+import logging
 
 from io import open
-
-
-def eprint(*args, **kwargs):
-    """Print to stderr."""
-    print(*args, file=sys.stderr, **kwargs)
 
 
 class FullDownload:
@@ -32,24 +28,26 @@ class FullDownload:
 
         subparser.add_argument(
             "--name",
-            help="Name you're going to assign to the bundle/file you're downloading."
+            help="Name of the folder you're storing the bundle/file in (Your name for the bundle).\
+                Defaults to the uuid of the bundle or file."
         )
 
         subparser.add_argument(
             "--replica",
-            help="Which cloud to download from first. One of 'aws', 'gc', or 'azure'.",
+            help="Which cloud to download from first. One of 'aws', 'gcp', or 'azure'.",
+            choices=['aws', 'gcp', 'azure'],
             default="aws"
         )
 
     @classmethod
     def _download_files(cls, files, folder, api, replica):
-        """Use the API class to make a put-files request on each of these files."""
-        for file in files:
+        """Use the API class to make a get-files request on each of these files."""
+        for file_ in files:
 
-            file_uuid = file[cls.UUID]
-            filename = file.get("name", file_uuid)
+            file_uuid = file_[cls.UUID]
+            filename = file_.get("name", file_uuid)
 
-            eprint("File {}: Retrieving...".format(filename))
+            logging.info("File {}: Retrieving...".format(filename))
 
             request = [
                 "get-files",
@@ -60,17 +58,21 @@ class FullDownload:
             response = api.make_request(request, stream=True)
 
             if response.ok:
-                eprint("File {}: GET successful. Writing to disk.".format(filename, file_uuid))
-                with open(folder + filename, "wb") as fh:
-                    fh.write(response.content)
+                file_path = os.path.join(folder, filename)
+                logging.info("File {}: GET successful. Writing to disk.".format(filename, file_uuid))
+                with open(file_path, "wb") as fh:
+                    # Process taken from
+                    # https://stackoverflow.com/questions/16694907/how-to-download-large-file-in-python-with-requests-py
+                    for chunk in response.iter_content(chunk_size=1024):
+                        if chunk:
+                            fh.write(chunk)
                 response.close()
-                eprint("File {}: write successful. Stored at {}.".format(filename, folder + filename))
+                logging.info("File {}: GET SUCCEEDED. Stored at {}.".format(filename, file_path))
 
             else:
-                eprint("File {}: GET FAILED. This uuid is neither a bundle nor a file.".format(filename))
-                eprint(response.text)
+                logging.info("File {}: GET FAILED. This uuid is neither a bundle nor a file.".format(filename))
+                logging.info(response.text)
                 response.close()
-                response.raise_for_status()
 
         return {"completed": True}
 
@@ -81,7 +83,7 @@ class FullDownload:
         bundle_name = args.get("name", bundle_uuid)
         replica = args["replica"]
 
-        eprint("Bundle {}: Retrieving...".format(bundle_uuid))
+        logging.info("Bundle {}: Retrieving...".format(bundle_uuid))
         request = [
             "get-bundles",
             bundle_uuid,
@@ -94,19 +96,16 @@ class FullDownload:
 
         if response.ok:
             files = response.json()["bundle"]["files"]
-            folder = bundle_name + "/"
+            folder = bundle_name
             if not os.path.isdir(folder):
                 os.mkdir(folder)
-            eprint("Bundle {}: GET successful. {} files to download".format(bundle_uuid, len(files)))
+                logging.info("Bundle {}: GET SUCCEEDED. {} files to download".format(bundle_uuid, len(files)))
+            logging.info("Request response:")
+            logging.info("{}\n".format(response.content.decode()))
 
         else:
-            eprint("Bundle {}: GET FAILED. Checking if uuid is a file.".format(bundle_uuid))
-            eprint(response.text)
-            response.close()
-            response.raise_for_status()
-
-        eprint("Request response:")
-        eprint("{}\n".format(response.content.decode()))
+            logging.info("Bundle {}: GET FAILED. Checking if uuid is a file.".format(bundle_uuid))
+            logging.info(response.text)
         return files, folder
 
     @classmethod
--- a/hca/full_download.py
+++ b/hca/full_download.py
@@ -1,0 +1,118 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import os
+import sys
+
+from io import open
+
+
+def eprint(*args, **kwargs):
+    """Print to stderr."""
+    print(*args, file=sys.stderr, **kwargs)
+
+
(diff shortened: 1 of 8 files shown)


hca/full_download.py
```diff
@@ -0,0 +1,118 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import os
+import sys
+
+from io import open
+
+
+def eprint(*args, **kwargs):
```
ttung: you should use https://docs.python.org/3/library/logging.html instead.

hca/full_download.py
```diff
@@ -0,0 +1,118 @@
...
+
+        subparser.add_argument(
+            "--replica",
+            help="Which cloud to download from first. One of 'aws', 'gc', or 'azure'.",
+            default="aws"
+        )
+
+    @classmethod
+    def _download_files(cls, files, folder, api, replica):
+        """Use the API class to make a put-files request on each of these files."""
```
ttung: copy-pasta

hca/full_download.py
```diff
@@ -0,0 +1,118 @@
...
+        for file in files:
```
ttung: avoid `file` since that's a keyword.

hca/full_download.py
```diff
@@ -0,0 +1,118 @@
...
+                "get-files",
+                file_uuid,
+                "--replica", replica,
+            ]
+
+            response = api.make_request(request, stream=True)
+
+            if response.ok:
+                eprint("File {}: GET successful. Writing to disk.".format(filename, file_uuid))
+                with open(folder + filename, "wb") as fh:
```
ttung: is this approach compatible with `stream=True`?  my guess is no.  `response.content` probably pulls this all into a single byte string.

hca/full_download.py
```diff
@@ -0,0 +1,118 @@
...
+                    fh.write(response.content)
+                response.close()
+                eprint("File {}: write successful. Stored at {}.".format(filename, folder + filename))
+
+            else:
+                eprint("File {}: GET FAILED. This uuid is neither a bundle nor a file.".format(filename))
```
ttung: super minor nit:

GET successful
GET FAILED

pick a case :)

hca/full_download.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+
+        subparser.add_argument(
+            cls.UUID,
+            help="UUID of the folder or file you want to download."
+        )
+
+        subparser.add_argument(
+            "--name",
+            help="Name you're going to assign to the bundle/file you're downloading."
```
ttung: this is somewhat confusing, though I'm not sure what's better. :(

Also, you should mention that it defaults to the uuid.

hca/full_download.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+
+        subparser.add_argument(
+            "--replica",
+            help="Which cloud to download from first. One of 'aws', 'gc', or 'azure'.",
```
ttung: now it's aws, gcp, or azure.

ttung: also see https://docs.python.org/3/library/argparse.html#choices

hca/full_download.py
```diff
@@ -0,0 +1,118 @@
...
+
+        files = [args]
+        folder = ""
+        response = api.make_request(request)
+
+        if response.ok:
+            files = response.json()["bundle"]["files"]
+            folder = bundle_name + "/"
+            if not os.path.isdir(folder):
+                os.mkdir(folder)
```
ttung: consider https://docs.python.org/2/library/os.html#os.makedirs

hca/full_download.py
ttung: also, you should do os.path.join(folder, filename)

hca/full_download.py
```diff
@@ -0,0 +1,118 @@
...
+            "--replica", replica,
+        ]
+
+        files = [args]
+        folder = ""
+        response = api.make_request(request)
+
+        if response.ok:
+            files = response.json()["bundle"]["files"]
+            folder = bundle_name + "/"
```
ttung: if you do the os.path.join above, then you don't actually need to stick on the / here.

but if you do, you should be using https://docs.python.org/2/library/os.html#os.pathsep

hca/full_upload.py
```diff
@@ -0,0 +1,177 @@
...
+
+        subparser.add_argument(
+            cls.FILE_OR_DIR_ARGNAME,
+            nargs="+",
+            help="Relative or direct path to folder with all bundle files."
+        )
+
+        subparser.add_argument(
+            cls.REPLICA_ARGNAME,
+            help="Which cloud to upload to first. One of 'aws', 'gc', or 'azure'.",
```
ttung: same as above.

hca/full_upload.py
```diff
@@ -0,0 +1,177 @@
...
+            default="aws"
+        )
+        subparser.add_argument(
+            "--staging-bucket",
+            help="Bucket within replica to upload to.",
+            default="hca-dcp-staging-test")
```
ttung: not sure you want to have a default value since most users would not be able to take advantage of it.

hca/full_upload.py
```diff
@@ -0,0 +1,177 @@
...
+
+import os
+import pprint
+import sys
+import uuid
+
+from io import open
+
+from .constants import Constants
+from .upload_to_cloud import upload_to_cloud
```
ttung: #5 has to land before this, no?

hca/full_upload.py
```diff
@@ -0,0 +1,177 @@
...
+                    full_file_name = os.path.join(path, filename)
+                    filenames.append(filename)
+                    files_to_upload.append(open(full_file_name, "rb"))
+            else:  # It's a file
+                filenames.append(os.path.basename(path))
+                files_to_upload.append(open(path, "rb"))
+
+        # Print to stderr, upload the files to s3 and return a list of tuples: (filename, filekey)
+        eprint("Uploading the following keys to aws:")
+        for file in filenames:
```
ttung: avoid `file`

hca/full_upload.py
```diff
@@ -0,0 +1,177 @@
...
+            eprint("\t", file)
+
+        uploaded_keys = upload_to_cloud(files_to_upload, args["staging_bucket"], args["replica"])
+
+        filename_key_list = list(zip(filenames, uploaded_keys))
```
ttung: i don't think you need to call `list(..)` here

hca/full_upload.py
```diff
@@ -0,0 +1,177 @@
...
+        for filename, key in filename_key_list:
+            eprint('\t{:<12}  {:<12}'.format(filename, key))
+        return filename_key_list
+
+    @classmethod
+    def _put_files(cls, filename_key_list, staging_bucket, api):
+        """Use the API class to make a put-files request on each of these files."""
+        bundle_uuid = str(uuid.uuid4())
+        files = []
+        for (filename, key) in filename_key_list:
```
ttung: i believe the standard python convention has no parens here (even though I actually like this formatting better myself).

hca/full_upload.py
```diff
@@ -0,0 +1,177 @@
...
+                response.raise_for_status()
+
+            eprint("Request response")
+            eprint("{}".format(response.content.decode()))
+        return bundle_uuid, files
+
+    @classmethod
+    def _put_bundle(cls, bundle_uuid, files, api):
+        """Use the API class to make a put-bundles request."""
+        file_args = [Constants.OBJECT_SPLITTER.join(["True", file["name"], file["uuid"], file["version"]]) for file in files]
```
ttung: split this across lines.

hca/full_upload.py
```diff
@@ -0,0 +1,177 @@
...
+        creator_uid = os.environ.get(cls.CREATOR_ID_ENVIRONMENT_VARIABLE, "1")
+        replica = "aws"
```
ttung: shouldn't this be from your command line parser?

hca/full_download.py
Mackey22: Will change. Assuming I should use logger.info(msg)?

hca/full_download.py
Mackey22: Adding uuid.

hca/full_download.py
Mackey22: Updating to limit choices to aws, gcp, or azure.

hca/full_download.py
Mackey22: Whoops lol switching

hca/full_download.py
Mackey22: Planning on changing to `file_`. Would it be better to change to a different name altogether?

hca/full_download.py
Mackey22: You're right. Have to iterate using` response.iter_content()`

Mackey22: 👍

hca/full_download.py
Mackey22: 👍

hca/full_download.py
Mackey22: 👍

hca/full_download.py
Mackey22: Don't see much difference between the two. `mkdir` makes all intermediate folders required, and they both raise an error if the folder already exists. Do you think it should raise an error if the folder already exists?

hca/full_upload.py
Mackey22: Yeah, didn't realize that until after. We can merge that one in first.

hca/full_upload.py
Mackey22: 👍

hca/full_upload.py
Mackey22: Fair. Was mostly for my convenience lol

hca/full_upload.py
Mackey22: Same question as above

hca/full_upload.py
Mackey22: Yeah you're right. I think I only called it because I wanted to print and see what was in the zip at some point.

hca/full_upload.py
Mackey22: Changing

hca/full_upload.py
Mackey22: Will do

hca/full_upload.py
Mackey22: Yeah definitely. Noticed and changed this in download, forgot to do the same for this one.

hca/full_download.py
ttung: Yeah, I would just use a different variable name.

I usually like using fn/fname (filename), fp/fpath (filepath), or fh (file handle), depending on what the type is.

hca/full_download.py
ttung: `os.mkdir` should _not_ make all the intermediate folders.

Mackey22:
@ttung @kislyuk ready for another look when you get the chance.

kislyuk:
@Mackey22 this PR seems to include some of the same diffs as the other one that I post-commit-reviewed, please check. It looks good to me otherwise.

Mackey22:
@kislyuk I merged the changes in to this branch from the previous branch before merging the previous branch into master, so there are some of the same diffs. I'll revert that merge then merge this into master. I should merge this one before #5 ?

kislyuk:
Sounds good.

hca/full_download.py
Mackey22: Whoops changed

PR #5, opened 2017-06-30
Installing the dependencies and file structure needed for high-level …

Mackey22:
…commands to actually put files into the cloud.


hca/upload_to_cloud.py
```diff
@@ -0,0 +1,82 @@
...
+import argparse
+import logging
+import mimetypes
+import os
+import uuid
+
+import boto3
+from boto3.s3.transfer import TransferConfig
+
+from io import open
```
ttung: move this import up.

hca/upload_to_cloud.py
```diff
@@ -0,0 +1,82 @@
...
+            key_names.append(key_name)
+
+    return key_names
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("files", nargs="+", type=lambda f: open(f, "rb"))
+    parser.add_argument("--staging-bucket", default="hca-dss-test")
```
ttung: again, not sure having a default makes sense.

what might make sense is a default from the environment, though.

hca/upload_to_cloud.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def encode_tags(tags):
+    return [dict(Key=k, Value=v) for k, v in tags.items()]
+
+
+def _mime_type(filename):
+    try:
+        mimetypes_tuple = mimetypes.guess_type(filename)
+        last_non_none_mimetype = list(filter(lambda x: x, list(reversed(mimetypes_tuple))))[0]
```
ttung: do you need all the `list` calls?

i would just do this as a for loop tbh.

hca/upload_to_cloud.py
Mackey22: 👍

hca/upload_to_cloud.py
Mackey22: Making it required and adding default to be os.getenv('STAGING_BUCKET')

hca/upload_to_cloud.py
Mackey22: This was much more complicated than it had to be.

Mackey22:
@ttung Tests went away because the arguments didn't line up with the function stubs. Was thinking about changing tests to reflect that but the tests are re-added in their proper form in the tests pr so I didn't worry about it.

hca/upload_to_cloud.py
```diff
@@ -0,0 +1,82 @@
...
+            s3.meta.client.put_object_tagging(Bucket=bucket.name,
+                                              Key=key_name,
+                                              Tagging=dict(TagSet=encode_tags(metadata))
+                                              )
+            key_names.append(key_name)
+
+    return key_names
+
+
+if __name__ == "__main__":
```
kislyuk: Is the ability to call this library file as a script intended for testing? If so, can you add a test that explicitly calls this as a script? Otherwise, it would be better to remove it.

PR #7, opened 2017-06-30
Auto-generate CLI from Swagger

Mackey22:


ttung:
I assume this PR will have a more descriptive name?

hca/__init__.py
```diff
@@ -1,14 +1,19 @@
...
+
 def main():
-    api = API("a", "b")
-    print("Hello HCA! :)")
-    return api.make_request(sys.argv[1:])
+    """Entrance to functionality."""
+    api = API()
+    response = api.make_request(sys.argv[1:])
+    if isinstance(response, requests.Response):
+        sys.stdout.write(response.content.decode())
```
ttung: You should always use `sys.stdout.write` _or_ `print`.  Not both.

hca/constants.py
```diff
@@ -0,0 +1,9 @@
+"""File to keep constants together."""
+
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+
+class Constants:
+    """Class to keep constants together."""
+
+    OBJECT_SPLITTER = "/"
```
ttung: This needs some explanation.

hca/define_api.py
```diff
@@ -1,37 +1,172 @@
...
+
+from .parser import get_parser, ADDED_COMMANDS
+from .constants import Constants
 
 
 class API:
     """Class for interacting with the REST Api."""
 
-    def __init__(self, api_url, user):
+    def __init__(self, test=False):
```
ttung: test is kind of ambiguous in intent.

hca/define_api.py
```diff
@@ -1,37 +1,172 @@
...
         """Initialize the CLI API."""
-        self.spec = self.get_spec()
-        self.base_url = self.spec['host'] + self.spec['basePath']
-        self.parser = get_parser(self.spec)
+        spec = self.get_spec(test)
```
ttung: since you're passing in `test` as a constructor argument, consider just storing it in `self.test`, and using that instead of passing it around.

same comment applies about the ambiguity of this name.

hca/define_api.py
```diff
@@ -1,37 +1,172 @@
...
+        scheme = spec["schemes"][0] if "schemes" in spec else "https"
```
ttung: inconsistent use of " vs '.

moved discussion to slack.

hca/define_api.py
```diff
@@ -1,37 +1,172 @@
...
+        self.base_url = "{}://{}{}".format(scheme, spec['host'], spec['basePath'])
```
ttung: consider https://docs.python.org/2/library/urlparse.html, though that might be an overkill.

hca/define_api.py
```diff
@@ -1,37 +1,172 @@
...
+    def get_spec(self, test=False):
+        """
+        Load the API specification.
+
+        :param test: boolean flag to indicate which spec to use.
+                     api_spec is the spec downloaded from the api on build.
+                     ../test/test is the mocked up spec I've been toying with.
+        :return:     The dictionary containing all swagger specification definitions.
+        """
+        url = os.path.dirname(os.path.realpath(__file__)) + "/api_spec.json"
```
ttung: `os.path.join`

hca/define_api.py
```diff
@@ -1,37 +1,172 @@
...
+        if test:
+            url = os.path.dirname(os.path.realpath(__file__)) + "/../test/test.json"
+
+        with open(url) as fp:
             api_spec_dict = json.load(fp)
         return api_spec_dict
 
+    def _build_url(self, endpoint, namespace):
+        """Add positional arguments into the url."""
+        split_endpoint = endpoint.split("-")[1:]
```
ttung: is this a convention?

consider doc-blocking this function.

hca/define_api.py
```diff
@@ -1,37 +1,172 @@
...
+
+        Arguments:
+         - arg: The value that is being assigned.
+         - payload: The payload structure before adding this value.
+         - hierarchy: Data structure that describes where in payload to put arg.
+        """
+        inner_payload = payload
+
+        # Iterating through payload structure to find the right value to set.
+        while len(hierarchy) > 1:
```
ttung: While this is probably ok given that hierarchy is not likely deep, you should just iterate through and find what you need and not take repeated slices of a list, which is actually fairly expensive.

hca/define_api.py
```diff
@@ -1,37 +1,172 @@
...
+        return url
+
+    def _add_arg(self, arg, payload, hierarchy):
+        """
+        Add a given argument to its associated payload.
+
+        Arguments:
+         - arg: The value that is being assigned.
+         - payload: The payload structure before adding this value.
+         - hierarchy: Data structure that describes where in payload to put arg.
```
ttung: Um... this is a bit ambiguous.

hca/define_api.py
```diff
@@ -1,37 +1,172 @@
...
+            if curr_level not in inner_payload:
+                inner_payload[curr_level] = [] if isinstance(hierarchy[1], list) else {}
+            inner_payload = inner_payload[curr_level]
+
+            hierarchy = hierarchy[1:]
+
+        curr_level = hierarchy[0]
+
+        # Command line argument represents a list of objects.
+        if type(curr_level) == list and len(curr_level) > 0:
```
ttung: typically `instanceof` is a safer thing to do, because `type(..)` requires an exact match of types.

hca/define_api.py
```diff
@@ -1,37 +1,172 @@
...
+                        value = json.loads(value)
+                    elif argument_type == "integer":
+                        value = int(value)
+                    elif argument_type == "number":
+                        value = float(value)
+                    elif argument_type == "boolean" and value == "True":
+                        value = True
+                    elif argument_type == "boolean" and value == "False":
+                        value = False
+                    obj[argument_name] = value
```
ttung:
```
else:
    raise ValueError("Unknown type")
```

hca/define_api.py
```diff
@@ -1,37 +1,172 @@
...
+        # Command line argument represents a list of standard types.
+        elif type(curr_level) == list:
+            for object_arg in arg:
+                inner_payload.append(object_arg)
+
+        # Command line argument represents a standard type.
+        else:
+            inner_payload[curr_level] = arg
+
+    def _build_payloads(self, endpoint, namespace):
```
ttung: this definitely needs a docblock.

setup.py
```diff
@@ -5,15 +5,18 @@
```
ttung: seems like a separate diff to me!

hca/__init__.py
Mackey22: 👍

hca/constants.py
Mackey22: Will add. It's the character we decided on to split the variables within an object when they're listed. So in the put-bundles command, the files should be specified as `<INDEXED><OBJECT_SPLITTER><NAME><OBJECT_SPLITTER><UUID><OBJECT_SPLITTER><VERSION>`

hca/define_api.py
Mackey22: Renaming to `use_test_api`

hca/define_api.py
Mackey22: 👍

hca/define_api.py
Mackey22: 👍

hca/define_api.py
Mackey22: I think I'll keep it as is b/c urlparse's urljoin doesn't add any functionality when scheme is not given as part of the base url.

hca/define_api.py
Mackey22: 👍

hca/define_api.py
Mackey22: doc-blocking? Will rethink this assumption with you tho

hca/define_api.py
Mackey22: Will clear this up.

hca/define_api.py
Mackey22: 👍

hca/define_api.py
Mackey22: 👍

hca/define_api.py
Mackey22: 👍

hca/define_api.py
Mackey22: 👍

setup.py
Mackey22: Agreed. I think I had to include it so the Travis build would pass.

Mackey22:
@ttung @kislyuk last thing needing review for this huge batch review is parser.py on this branch

hca/parser.py
```diff
@@ -15,36 +23,331 @@ def make_name(http_method, path_split):
...
+                inner_param["in"] = "body"
+                inner_param["name"] = "irrelevant"
+                inner_param['req'] = param['req']
+                param = inner_param
+
+            _recursive_indexing(spec, param, [], indexed_parameters)
     return indexed_parameters
 
 
+def create_help_statement(endpoint_info):
```
ttung: I looked at the circumstances this is generated, and it's kind of confusing why this is present, and how we expect users to understand this.

Do we expect that they know what the HTTP endpoints are, and that they are just using the CLI to drive operations?  Or do we expect them to learn about the operations from the help?  I think the latter makes more sense, and in that case, I don't really see the value of this.

hca/parser.py
```diff
@@ -1,8 +1,16 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import argparse
+import json
+import jsonpointer
```
Mackey22: 👍

hca/parser.py
```diff
@@ -15,36 +23,331 @@ def make_name(http_method, path_split):
     return name
 
 
-def index_parameters(endpoint_info):
+def _propogate_info(next_level, param):
```
Mackey22: 👍

hca/parser.py
```diff
@@ -15,36 +23,331 @@ def make_name(http_method, path_split):
...
+    """Helper to propogate these values into the next level of the data structure."""
+    next_level["name"] = param["name"]
+    next_level["req"] = param["req"]
+    next_level["in"] = param["in"]
+
+
+def _array_indexing(spec, param, hierarchy_clone, indexed_parameters):
+    next_level = param["items"]
+    _propogate_info(next_level, param)
```
Mackey22: 👍

hca/parser.py
```diff
@@ -15,36 +23,331 @@ def make_name(http_method, path_split):
...
+                           parsing. Basic format is:
+                           {
+                                <endpoint-name>: {
+                                    description: str,
+                                    options: {
+                                        <option-arg-name> {
+                                            array: bool,
+                                            description: str,
+                                            format: str,
+                                            hierarchy: list,
```
Mackey22: It's described fully in `define_api.py`. I'll add a short description next to this

setup.py
Mackey22: This pr relies on the added packages to properly run. I included the setup.py diff in the readme/infra pr, so when I committed this, I checked setup.py out from that branch so this would build.

hca/parser.py
Mackey22: Some arguments are required for certain paths and optional/ignored otherwise. I included this so users would know when these arguments are optional or required. For instance, in `hca get-bundles`, version and replica are required when uuid is given, but optional otherwise. The presentation can definitely be improved, but I think this is information that's going to be vital for users.

hca/define_api.py
```diff
@@ -1,37 +1,209 @@
...
 
     def get_spec(self):
-        """Load the API specification."""
-        with open(os.path.dirname(os.path.realpath(__file__)) + "/api_spec.json") as fp:
+        """
+        Load the API specification.
+
+        self.use_test_api is a boolean flag to indicate which spec to use.
+            api_spec is the spec downloaded from the api on build.
+            ../test/test is the mocked up spec I've been toying with.
```
Mackey22: Definitely a bit wonky (and looking through git history the way I had originally done it). I could call `os.path.dirname` again before, but I feel like that would be even messier - pretty much everyone understands what two dots means.

Mackey22: nvm fixing this

hca/parser.py
Mackey22: Removing unhelpful message saying what arguments are required for which endpoints. In future pr, will throw an error if subcommand given without proper query string args. For instance, if GET bundles/:uuid/:version doesn't provide a replica argument.

PR #10, opened 2017-07-05
Documentation setup with Sphinx.

Mackey22:
Setup initial Sphinx infrastructure.


kislyuk:
Very nice!

In general, I prefer to add docstrings to the methods/classes themselves:

```
class Foo:
    """
    This docstring will automatically be slurped into autodocs. It can use all sphinx and ReST markup.
    """
    def bar(self):
        """
        This is a method or function docstring.
        """
        # Do useful stuff. This is a comment, it won't be interpreted as a docstring.
        pass
```

Mackey22:
Ugh ideal scenario for a comment. That doc was giving me trouble before, but now it's not so I'll remove the `get_parser.__doc__` assignment

PR #16, opened 2017-07-12
Support uploading bundle from staged s3 bucket to DSS

Mackey22:
Now supports the command `hca upload s3://my-remote-bucket/fastqs/bundle`.
All files within the bundle must have all required checksums precomputed. These are: 
* sha1
* sha256
* crc32c
* s3_etag
* content-type

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,21 @@
 language: python
 cache: pip
-
 python:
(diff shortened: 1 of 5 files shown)


Mackey22:
Ignoring codecov complaints, merging, and packaging new PyPi release.

PR #19, opened 2017-07-13
Add AddedCommand superclass for all api endpoints.

Mackey22:
Most of this logic is just replayed from previous commits. I need this class to make it easy to generate the python bindings for the api-defined commands and higher-level commands that others will create. 

Api-defined commands will override `_get_base_url`, `_get_endpoint_name`, and `_get_endpoint_info` when they're auto-generated. Higher-level commands should override `add_parser`, `run_cli`, and `run` when they're implemented. 

The only significant difference in `run_cli` and `run` is that `run` asks for the entire body payload as one object rather than breaking down the individual parts for now. Not sure if we should keep it this way in the future. 

Travis changes were unnecessary because they're landed already. Hence the revert in https://github.com/HumanCellAtlas/data-store-cli/commit/52fbe2eba0da223b12455fa11fc8858ebb9e753a


hca/added_command.py
```diff
@@ -0,0 +1,278 @@
...
+    """Object to parse json objects when they're inputted in the cli."""
+
+    def __init__(self, option_strings, dest, nargs=None, **kwargs):
+        super(AddObject, self).__init__(option_strings, dest, **kwargs)
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        setattr(namespace, self.dest, json.loads(values))
+
+
+class AddedCommand:
```
ttung: python2 right?  should be:

`class AddedCommand(object):`

hca/added_command.py
```diff
@@ -0,0 +1,278 @@
...
+        endpoint_info = cls._get_endpoint_info()
+        for positional_arg in endpoint_info['positional']:
+            h = endpoint_info['description']
+
+            argtype = cls._get_arg_type(positional_arg['type'])
+
+            subparser.add_argument(
+                positional_arg['argument'],
+                nargs=None if positional_arg['required'] else "?",
+                help=h,
```
ttung: h not used elsewhere.  would just inline that here.

hca/added_command.py
```diff
@@ -0,0 +1,278 @@
...
+        for (optional_name, optional_data) in endpoint_info['options'].items():
+            h = endpoint_info['description']
+
+            argtype = cls._get_arg_type(optional_data['type'])
+            actiontype = cls._get_action(optional_data['type'])
+
+            optional_name_with_dashes = optional_name.replace("_", "-")
+            subparser.add_argument(
+                "--" + optional_name_with_dashes,
+                dest=optional_name,  # So we can send correctly formatted objects to api.
```
ttung: wild question: what if optional_name had a hyphen?

hca/added_command.py

hca/added_command.py

hca/added_command.py
Mackey22: An optional_name would only start with a hyphen if dss-api.yml includes parameters with hyphens in them. I'm failing with an APIException telling user to contact DCC to change the api in this case.

Mackey22: This change where this should take place should actually happen in branch jmackey-python-bindings-make-classes, which is stacked on top of this. Tracked by https://github.com/HumanCellAtlas/data-store-cli/issues/22

Mackey22:
Fixed comments. Merging unless you have any comments @kislyuk

Mackey22:
Codecov checks didn't pass because I'm planning on adding full functionality testing when the bindings are actually exposed.

PR #21, opened 2017-07-19
Python Bindings - full autogenerate from endpoint

Mackey22:
* `hca/api/__init__.py` contains the true python functions that will be available.
* Everything in `hca/api/autogenerated/` is classes autogenerated by the api spec.
* `api.jinja` is the template for class structure of every file in `hca/api/autogenerated`.
* `functions.jinja` is the template for the Python functions that will be exposed to the api in `hca/api/__init__.py`.
* Changes in `full_upload.py` and `upload_to_cloud.py` came from cherry-picking the changes made in `jmackey-upload-from-cloud` branch.
* Testing changes included adding a real bundle to upload, switching the variable `api` to `cli` in earlier test cases to avoid namespace collisions, and fully testing the python bindings.


hca/added_command.py
```diff
@@ -217,37 +218,71 @@ def _build_non_body_payloads(cls, namespace):
...
             # If the argument is positional, belongs in path, not a payload
             if arg_name not in endpoint_info["options"]:
                 continue
 
+            if arg_name in endpoint_info['body_params']:
+                # Check that it conforms to the right schema
+                try:
+                    jsonschema.validate(arg, endpoint_info['body_params'][arg_name])
+                except jsonschema.ValidationError as e:
+                    print(e)
```
ttung: is this supposed to be here?

hca/added_command.py
```diff
@@ -217,37 +218,71 @@ def _build_non_body_payloads(cls, namespace):
...
+        for i in range(num_positionals):
+            if i < len(args) and args[i] is not None:
+                query_route.append(args[i])
+
+        # for i in range(num_positionals, len(args))
+        optionals = {}
+        i = num_positionals
+        # Not using enumerate because I'm not sure it keeps order consistent when iterating.
+        # Iterating over dictionaries with the same keys is guaranteed to be consistent ordering.
+        for optional_name in sorted(endpoint_info['options'].keys()):
```
ttung: You can do:

`for ix, optional_name in enumerate(sorted(endpoint_info['options'].keys()):`

Any usages of `ix` would require `ix + num_positionals` though.

hca/added_command.py
```diff
@@ -217,37 +218,71 @@ def _build_non_body_payloads(cls, namespace):
...
 
+        split_endpoint = endpoint_name.split("-")[1]
         query_route = [split_endpoint]
-        query_route.extend(args)
+        num_positionals = len(endpoint_info['positional'])
+        for i in range(num_positionals):
+            if i < len(args) and args[i] is not None:
+                query_route.append(args[i])
+
+        # for i in range(num_positionals, len(args))
```
ttung: remove?

hca/api/templates/functions.jinja
```diff
@@ -0,0 +1,19 @@
+{% for class_info in classes %}
```
ttung: Is the scary message `"""This file is autogenerated according to HCA api spec. DON'T TOUCH!"""` supposed to be here?

hca/regenerate_api.py
```diff
@@ -320,39 +359,36 @@ def generate_python_bindings(test_api_path=None):
 
     # Blow away previous files in the autogenerated folder and remake that directory
     dirname = os.path.dirname(__file__)
+    api_function_path = os.path.join(dirname, "api", "__init__.py")
+    if os.path.exists(api_function_path):
+        os.remove(api_function_path)
```
ttung: https://docs.python.org/3/library/os.html#os.truncate

hca/regenerate_api.py
```diff
@@ -320,39 +359,36 @@ def generate_python_bindings(test_api_path=None):
...
     open(os.path.join(autogenerated_path, "__init__.py"), 'a').close()
 
-    for (endpoint_name, endpoint_info) in param_holders.items():
-
-        template_loader = FileSystemLoader(searchpath=os.path.join(dirname, "api", "templates"))
+    function_payload = {'classes': []}
 
-        # An environment provides the data necessary to read and parse our templates. Pass in the loader object here.
-        template_env = Environment(loader=template_loader)
+    for (endpoint_name, endpoint_info) in param_holders.items():
```
ttung: i think it's just 

`for endpoint_name, endpoint_info in param_holders.items():`

hca/upload_to_cloud.py
```diff
@@ -89,6 +89,8 @@ def upload_to_cloud(files, staging_bucket, replica, from_cloud=False):
                     "hca-dss-content-type": _mime_type(fh.raw.name)
                 }
 
+                raw_fh.close()
```
ttung: if this is correct, your docblock on L61 is inaccurate.

test/test.py
```diff
@@ -256,6 +261,38 @@ def test_upload_to_cloud_from_s3(self):
...
         pass
 
+    def test_python_bindings(self):
+        from hca import api
+        cli = hca.define_api.API()
+
+        bundle_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "bundle")
+        namespace = {'file_or_dir': [bundle_path],
+                     'replica': "aws",
+                     'staging_bucket': "org-humancellatlas-dss-test-jmackey"}
```
ttung: is this desirable?

test/test.py
```diff
@@ -256,6 +261,38 @@ def test_upload_to_cloud_from_s3(self):
...
+        bundle_output = hca.full_upload.FullUpload.run(namespace, cli)
+
+        # Test get-files and head-files
+        file_ = bundle_output['files'][0]
+        self.assertTrue(api.get_files().ok)
+        self.assertTrue(api.get_files(file_['uuid'], replica="aws").ok)
+        # self.assertTrue(api.head_files(file_['uuid']).ok)  # Think api service might not be operating for head_files
```
ttung: is this true?

hca/added_command.py
Mackey22: jsonschema validation or error printing?

Jsonschema validation is important because we want to make sure that objects being submitted through the python bindings are actually adhering to the format described in the swagger spec. 

Printing the exception probably isn't the best pattern, but a couple times it's helped me in debugging. I can probably find a better way to reformat that.

hca/added_command.py
Mackey22: Looks good. This implementation is actually slightly irrelevant though because in the next pr, the run method for each autogenerated command just takes everything as a dict, which is passed in using locals() from the python bindings. This saves the trouble of this somewhat convoluted section.

hca/api/templates/functions.jinja
Mackey22: No, I'm adding that manually in `regenerate_api.py` - line 440. There's an explanation there for why I add it there instead of here. Gist of it is there are two times I'm adding using this functions template, so to keep formatting stable even though I'm adding twice, I add the imports and the scary message manually to the top of this file.

hca/added_command.py

hca/regenerate_api.py
Mackey22: Truncate and ftruncate don't support Windows until v3.5. Given that we're focused on providing cross-platform support, sticking with these more base methods might be better. Thoughts?

hca/regenerate_api.py
Mackey22: 👍

hca/regenerate_api.py
ttung: :(

That's kind of unfortunate, but makes sense.

hca/added_command.py
ttung: I meant the print.

ttung: I think you might be able to chain the exception.

`raise ValueError("blablah", ex)`

hca/api/templates/functions.jinja
ttung: Got it.

test/test.py
Mackey22: Yeah...
```
In [1]: from hca.api import head_files

In [2]: x = head_files("322442dd-22a1-4806-902f-5b781f8dcdf8")

In [3]: x.json()
---------------------------------------------------------------------------
JSONDecodeError                           Traceback (most recent call last)
<ipython-input-3-e89c9884c3f9> in <module>()
----> 1 x.json()

~/Documents/dss-cli/v3a/lib/python3.6/site-packages/requests/models.py in json(self, **kwargs)
    884                     # used.
    885                     pass
--> 886         return complexjson.loads(self.text, **kwargs)
    887 
    888     @property

/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/__init__.py in loads(s, encoding, cls, object_hook, parse_float, parse_int, parse_constant, object_pairs_hook, **kw)
    352             parse_int is None and parse_float is None and
    353             parse_constant is None and object_pairs_hook is None and not kw):
--> 354         return _default_decoder.decode(s)
    355     if cls is None:
    356         cls = JSONDecoder

/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/decoder.py in decode(self, s, _w)
    337 
    338         """
--> 339         obj, end = self.raw_decode(s, idx=_w(s, 0).end())
    340         end = _w(s, end).end()
    341         if end != len(s):

/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/decoder.py in raw_decode(self, s, idx)
    355             obj, end = self.scan_once(s, idx)
    356         except StopIteration as err:
--> 357             raise JSONDecodeError("Expecting value", s, err.value) from None
    358         return obj, end

JSONDecodeError: Expecting value: line 1 column 1 (char 0)

In [4]: x.content
Out[4]: b''

In [5]: x.status_code
Out[5]: 200
```

hca/added_command.py

hca/upload_to_cloud.py
Mackey22: Not sure I follow. It's an open binary file that needs to be closed when it's done being used. What discrepancy are you seeing?

test/test.py
Mackey22: Maybe another staging bucket:  `org-humancellatlas-dss-cli-test`?

hca/upload_to_cloud.py
ttung: It's a list of binary files, but I would more accurately characterize it as a list of binary file handles.  Binary files could be read as either binary file names or binary file handles.  I had to trace the caller to see how it was managed.

test/test.py
ttung: I think this is correct, unless the content-type is set to application/json.

Mackey22: Got it

hca/upload_to_cloud.py
Mackey22: Sounds good. I'll update that language

hca/added_command.py
Mackey22: wrt the above, going to leave this as is unless you think that's inappropriate

hca/upload_to_cloud.py
kislyuk: I would go further and say that I don't understand why this function is closing the file handles. Closing the file handles should be the caller's concern, it shouldn't incur this side effect.

Mackey22: 👍  https://github.com/HumanCellAtlas/data-store-cli/pull/21/commits/19d2ed0daae01bfb30790cdae222b5378571f6f2

Mackey22:
@kislyuk In the next diff, it's being renamed to `upload`. There will be a requirement that the class names, file names, and command names line up for the higher-level commands. I can rename now but in the next diff I did a git mv from `hca/full_upload.py` to `hca/api/abstractions/full_upload.py`. Will renaming full_upload something more descriptive complicate that?

kislyuk:
`upload` sounds good, thanks. No issue with the resulting strategy.

Mackey22:
Ok, great. You mind approving then @kislyuk ?

PR #24, opened 2017-08-07
[EASY] Upload real bundle data instead of random files.

Mackey22:
Uploading real data during testing. Taking this out of #21


Mackey22:
@ttung This good to merge then?

PR #25, opened 2017-08-07
Jmackey python bindings abstract functions autogenerated

Mackey22:
Separate pr changes to the auto-generated python bindings functions to avoid too many changes in one pr. 

Biggest changes: 
1. Autogenerated warning at top of every file. 
2. Abstractions import. 
3. Conversion from ordering args to sending args as dict from locals()
4. Additional file - hca/api/abstractions/__init__.py - necessary to avoid circular imports b/c the abstract functions depend on the autogenerated functions. So just regenerate the same functions in the abstractions directory.


Mackey22:
Closing this pr b/c the code used to create the autogenerated bindings is not in here. Will just include the new python bindings in the next pr.

PR #28, opened 2017-08-10
cli integration

Mackey22:
Integrate the cli on top of the python bindings and test that functionality. Again, there's baseline auth here, but the real version will be integrated in the diff on top of this.


hca/cli.py
```diff
@@ -0,0 +1,65 @@
...
+        """Function to actually make request to api."""
+        if not args:
+            self.parser.print_help()
+            self.parser.exit(1)
+        namespace = self.parse_args(args)
+        endpoint = args[0]
+
+        command = self.commands.get(endpoint, None)
+        if command:
+            return command.run_cli(namespace)
```
kislyuk: Why do you need two distinct entry points for the CLI and the bindings?

Nomenclature: `run_cli()` and `namespace` are confusing. `run_from_cli()` and `args` might work better.

PR #29, opened 2017-08-14
Jmackey popup auth

Mackey22:
* New auth-flow in added_command.py. First checks if access_token added as input. Next checks config. Next, checks for service account. Otherwise throws an error. Didn't provide popup here because gives very clear directions to use `hca login`. Also, don't save credentials to config file for service accounts - though that could be pretty simple to implement. 
* Reconfigure request sending to respond to 401 responses with a refreshed token if the endpoint requires auth. 
* Setup `hca login` command. If access token is provided, save that (After that token expires, though, you can't do anything with it). Elif in a terminal, oauth flow sign-in. 
* Only check credentials if security is required by the endpoint. 
* Testing w/ contextmanager for config files.


Makefile
```diff
@@ -27,6 +27,7 @@ install:
 	-rm -rf dist
 	python setup.py bdist_wheel
 	pip install --upgrade dist/*.whl
+	$(MAKE) bindings
```
kislyuk: This is still not good. From a clean checkout, this will result in a broken installed package. Let's chat about this offline.

hca/added_command.py
```diff
@@ -207,27 +212,103 @@ def _build_body_payload(cls, namespace):
...
+                    client_secret=config.client_secret,
+                    scopes=["https://www.googleapis.com/auth/userinfo.email"],
+                    refresh_token=config.refresh_token,
+                    token_expiry=None,
+                    token_uri=config.token_uri,
+                    user_agent=None,
+                    revoke_uri=None
+                )
+                # Silence ResourceWarning from httplib2 socket being open for connection pooling.
+                warnings.simplefilter("ignore")
```
kislyuk: This is not an appropriate command for a library to run. Let's figure out a way to go without it.

hca/added_command.py
```diff
@@ -207,27 +212,103 @@ def _build_body_payload(cls, namespace):
...
+                raise ValueError("The access token you've supplied in the kwargs is not valid."
+                                 " Please refresh and try again. You may also run `hca login`"
+                                 " to get a more permanent hca configuration.")
+            else:
+                logging.info("Found access token in kwargs.")
+                access_token = args['access_token']
+
+        # Checking config
+        elif config.get('access_token', None):
+            logging.info(config['access_token'])
```
kislyuk: Do not log credentials!

Mackey22: 👍  vestige from debugging

setup.py
```diff
@@ -16,14 +16,16 @@
         "six==1.10.0"
     ],
     install_requires=[
```
Mackey22: Alphabetizing dependencies, removing google-auth and google-api-python-client, and adding six

Makefile
ttung: Wild idea: can travis commit into the repo?  We should have a travis cronjob build the bindings, and then commit them if there's any changes. :)

hca/added_command.py
```diff
@@ -207,27 +212,103 @@ def _build_body_payload(cls, namespace):
...
+    @classmethod
+    def _get_access_token(cls, args, retry):
+        config = Config(Constants.TWEAK_PROJECT_NAME)
+        access_token = None
+
+        # kwargs access_token input
+        if 'access_token' in args:
+            if retry:
+                logging.info("Access token taken from kwargs invalid.")
+                raise ValueError("The access token you've supplied in the kwargs is not valid."
```
Mackey22: Should I be avoiding words like kwargs? I feel like other than that, it gives the user a pretty descriptive reason for why authentication failed (description continues below). Maybe I should add `"Authentication failed because the ..."`.

hca/added_command.py
```diff
@@ -207,27 +212,103 @@ def _build_body_payload(cls, namespace):
...
+                                 " Please refresh and try again. You may also run `hca login`"
+                                 " to get a more permanent hca configuration.")
+            else:
+                logging.info("Found access token in kwargs.")
+                access_token = args['access_token']
+
+        # Checking config
+        elif config.get('access_token', None):
```
Mackey22: In the test contextmanager, I tried to reset config by using something along the lines of `del config['access_token']`, but that didn't actually remove that variable from the config. So instead, I set them to `None` after I was done using them. This is a result of that. The only other solution I see around this is changing Tweak. @kislyuk am I misusing this?

hca/added_command.py
```diff
@@ -293,23 +373,28 @@ def run(cls, args):
...
+
+        requires_auth = cls._get_endpoint_info().get('requires_auth', False)
+
+        if requires_auth:
+            header_payload.update(cls._get_auth_header(args))
+
+        response = requests.request(**request_args)
+
+        # Maybe auth didn't work. Refresh token and try again.
+        if response.status_code == 401 and requires_auth:
```

hca/added_command.py
```diff
@@ -207,27 +212,103 @@ def _build_body_payload(cls, namespace):
...
+                         " Grabbing service account credentials from there.")
+            service_account_credentials_filename = os.environ['GOOGLE_APPLICATION_CREDENTIALS']
+
+            if not os.path.isfile(service_account_credentials_filename):
+                raise EnvironmentError(
+                    "File {} (pointed by GOOGLE_APPLICATION_CREDENTIALS"
+                    " environment variable) does not exist!".format(service_account_credentials_filename))
+
+            service_account_class = oauth2client.service_account.ServiceAccountCredentials
+            service_account_credentials = service_account_class.from_json_keyfile_name(
```
Mackey22: Using the google package directly now so this isn't as relevant, but I had seen setting a long class name this way in the argparse source code so I thought that could be nicer than wrapping the filename underneath. 

If the file is broken now the google auth package will handle it.

hca/api/composite_commands/login.py
```diff
@@ -0,0 +1,181 @@
...
+            client_secrets_file_path,
+            scope="https://www.googleapis.com/auth/userinfo.email",
+            message=oauth2client.tools.message_if_missing(client_secrets_file_path)
+        )
+
+        success = False
+        port_number = 0
+
+        auth_host_name = "localhost"
+        auth_host_port = [8080, 8090]
```
Mackey22: Now using the google flow pkg so not our concern anymore.

hca/added_command.py
Mackey22: Changed

hca/added_command.py
Mackey22: Per usage of Tweak, keeping this as is.

PR #30, opened 2017-08-14
Get argparser to return help string instead of printing

Mackey22:
The slack-bot should be able to get help in much the same way that the cli does, but the standard argparse ArgumentParser prints any help messages immediately and exits. This overrides the printing functions to throw a PrintingException whenever a print function would have been called. The system then catches these exceptions and returns the embedded message in string format so the cli will have no change in behavior, but the messages can be sent to the slackbot.


kislyuk:
Why not just wrap the whole thing in `try: ... except SystemExit as e: ...`?

Mackey22:
The help/error message is printed and not included in the SystemExit.

hca/argument_parser.py
```diff
@@ -0,0 +1,16 @@
+import argparse
+
+from .error import PrintingException
+
+
+class NonPrintingParser(argparse.ArgumentParser):
+    """ArgumentParser that will return help and error strings instead of printing."""
+
+    def __init__(self, *args, **kwargs):
+        super(NonPrintingParser, self).__init__(*args, **kwargs)
```
kislyuk: Doesn't seem necessary

hca/cli.py
```diff
@@ -1,20 +1,21 @@
...
+from .error import PrintingException
 
 
 class CLI:
     """Class for interacting with users through a CLI."""
 
     def __init__(self, test_api_path=None):
         """Initialize the CLI API."""
-        self.parser = argparse.ArgumentParser()  # Need to add description based on spec.
+        self.parser = NonPrintingParser()
```
kislyuk: Comment about adding docs from swagger is still useful, let's leave it in.

hca/cli.py
```diff
@@ -46,9 +47,15 @@ def parse_args(self, args):
     def make_request(self, args, stream=False):
         """Function to actually make request to api."""
         if not args:
-            self.parser.print_help()
-            self.parser.exit(1)
-        namespace = self.parse_args(args)
+            return self.parser.format_help()
+
+        # Raises PrintingException if Base ArgumentParser was going to print.
```
kislyuk: s/Raises/Handles?

hca/argument_parser.py
Mackey22: 👍

hca/cli.py
Mackey22: 👍

hca/cli.py
Mackey22: 👍

Mackey22:
Responded to comments indicated. Waiting until `jmackey-cli-integration` is merged so this can merge right into master and keep the other branch clean.

PR #32, opened 2017-08-14
Stream print & print bytes

Mackey22:
* This handles data formats that the cli previously didn't decode (specifically gzip). 
* Also handles the case where you call get-files on an enormous file. Streams the content instead of loading it immediately. 
* Changes the `make test` target to include install, which pip installs the package and puts hca into the namespace. I couldn't think of another way to check that the output printed bytes - open to other suggestions if you think this is a bad solution.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ lint:
 hca/api_spec.json:
 	curl https://hca-dss.czi.technology/v1/swagger.json > hca/api_spec.json
 
-test: lint bindings
+test: lint install
 	pip install six coverage
 	coverage run --source=$$(python setup.py --name) ./test/test.py
 
--- a/hca/__init__.py
+++ b/hca/__init__.py
@@ -26,17 +26,13 @@ def main():
     if isinstance(response, requests.Response):
         for chunk in response.iter_content(chunk_size=Constants.CHUNK_SIZE, decode_unicode=True):
             if chunk:  # filter out keep-alive new chunks
-                _write_binary_to_stdout(chunk)
+                if six.PY3:
+                    sys.stdout.buffer.write(chunk)
+                else:
+                    sys.stdout.write(chunk)  
     elif isinstance(response, dict):
         print(json.dumps(response))
     elif isinstance(response, str):
         print(response)
     else:  # Unicode
         print(response.decode())
-
-
-def _write_binary_to_stdout(content):
-    if six.PY3:
-        sys.stdout.buffer.write(content)
-    else:
-        sys.stdout.write(content)
--- a/hca/__init__.py
+++ b/hca/__init__.py
@@ -3,8 +3,11 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import json
+import os
 import requests
 import sys
+
+import six
 
 from .constants import Constants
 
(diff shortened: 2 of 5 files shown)


hca/__init__.py
```diff
@@ -6,13 +6,21 @@
...
     response = cli.make_request(sys.argv[1:])
     if isinstance(response, requests.Response):
-        print(response.content.decode())
+
+        for chunk in response.iter_content(chunk_size=Constants.CHUNK_SIZE, decode_unicode=True):
+            if chunk:  # filter out keep-alive new chunks
+                try:
+                    print(chunk.decode())
+                except UnicodeDecodeError:
+                    print(chunk)
```
kislyuk: Nope, don't decode!

Mackey22: Resolved

ttung:
This looks good.  Does travis have a mechanism by which we can do CI on windows?

Makefile
```diff
@@ -9,7 +9,7 @@ lint:
 hca/api_spec.json:
 	curl https://hca-dss.czi.technology/v1/swagger.json > hca/api_spec.json
 
-test: lint bindings
+test: lint install
```
ttung: why this change?  test doesn't regenerate bindings?

hca/__init__.py
```diff
@@ -3,16 +3,36 @@
...
     else:
         print(json.dumps(response))
+
+
+def _write_binary_to_stdout(content):
+    if six.PY3:
+        sys.stdout.buffer.write(content)
+    else:
+        # Windows includes carriage returns
+        if sys.platform == 'win32':
```
ttung: wouldn't it make sense to do this once before your for loop?

test/test.py
```diff
@@ -22,6 +22,8 @@
 from hca import constants  # noqa
 from test.config import override_oauth_config  # noqa
 
+dirpath = os.path.dirname(os.path.realpath(__file__))
```
ttung: why not move this just to `test_get_files_cli`?

hca/__init__.py
Mackey22: True. Will change

test/test.py
Mackey22: Was going to refactor for all tests using `dirpath` to just use this but didn't want to clutter the pr. Will just move to `test_get_files_cli` for now.

Makefile
Mackey22: Tests now rely on calling `hca get-files` cmd in a subprocess, so that command has to be installed. Thoughts on this pattern overall?

ttung: Oh I see.  Install relies on bindings.  Carry one.

ttung:
This diff is much larger than it was maybe an hour ago when I saw it last, FYI.

Mackey22:
Yeah, the branch underneath was rebased - need to rebase this one.

Mackey22:
Can't easily do CI on Windows: https://github.com/travis-ci/travis-ci/issues/216

Ready for review.

hca/__init__.py
```diff
@@ -3,16 +3,36 @@
...
     if isinstance(response, requests.Response):
-        print(response.content.decode())
+        for chunk in response.iter_content(chunk_size=Constants.CHUNK_SIZE, decode_unicode=True):
+            if chunk:  # filter out keep-alive new chunks
+                _write_binary_to_stdout(chunk)
     else:
         print(json.dumps(response))
+
+
+def _write_binary_to_stdout(content):
```
ttung: At this point, you might as well inline this method.

test/test.py
```diff
@@ -158,6 +158,26 @@ def test_index_parameters(self):
...
+        args = {'file_or_dir': [file_path],
+                'replica': "aws",
+                'staging_bucket': "org-humancellatlas-dss-cli-test"}
+        response = api.upload(**args)
+
+        file_uuid = response['files'][0]['uuid']
+
+        process = subprocess.Popen(["hca", "get-files", file_uuid, "--replica", "aws"], stdout=subprocess.PIPE)
+        out, _ = process.communicate()
+        self.assertIsInstance(out, bytes)
```
ttung: Also, do you want to verify that the output is correct?

ttung:
See comments pls.

hca/__init__.py
Mackey22: 👍

test/test.py
Mackey22: 👍 https://github.com/HumanCellAtlas/data-store-cli/pull/32/commits/9f9dc2d2d33a0ccfc4d22ce348893725baf8973e

PR #35, opened 2017-08-15
[EASY] Remove setting logging status to logging.info

Mackey22:


kislyuk:
Thanks, in general the library should not be configuring logging (especially not on import), but the CLI entry point should.

PR #41, opened 2017-08-22
use module-level imports to break circular dependencies.

### LINKED ISSUE #39: Dedup templates and eliminate circular imports
Currently the API methods are generated twice, once in api and once in api.composite_commands. Eliminate the generation in composite_commands. Fix the imports in composite_commands so they don't result in circular import errors.

ttung:
Resolves #39


ttung:
There's some issues with python < 3.5. :(

ttung:
ok ready for review.

Issue #40, opened 2017-08-22
implement an async uploader for the cli


kozbo:

ttung:
Done!

PR #44, opened 2017-08-28
Support async upload.

ttung:
1. If the server returns 202, we poll (with a timeout & backoff).
2. Add test.

Addresses #40 
connects to #40


ttung:
Updated with `BACKOFF_FACTOR`.

PR #46, opened 2017-08-28
Clean up logging.

ttung:
Stop writing to sys.stderr and use python's logging module instead.

### DIFF
--- a/hca/api/composite_commands/download.py
+++ b/hca/api/composite_commands/download.py
@@ -1,10 +1,10 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import os
-import sys
 
 import hca.api
 from io import open
+from ... import infra
 from ...added_command import AddedCommand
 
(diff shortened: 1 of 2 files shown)


hca/api/composite_commands/download.py
```diff
@@ -45,42 +45,45 @@ def _get_endpoint_info(cls):
...
     def _download_files(cls, files, folder, replica):
         """Use the python bindings to make a get-files request on each of these files."""
+        logger = infra.get_logger(Download)
         for file_ in files:
 
             file_uuid = file_['uuid']
             filename = file_.get("name", file_uuid)
 
-            sys.stderr.write("\nFile {}: Retrieving...".format(filename))
+            logger.info("File {}: Retrieving...".format(filename))
```
Bento007: Why not use the same format as dss-datastore `logger("%s", f"...")`?

PR #59, opened 2017-09-27
Refactor to support multiple command groups

sampierson:
Some cleanup.
Some foundational refactoring.


test/test_dss_cli.py
```diff
@@ -172,24 +172,26 @@ def test_get_files_cli(self):
...
         """Test that the parser parses arguments correctly."""
 
         cli = hca.dss.cli.CLI()
 
         args = ["put-files", "134", "--bundle-uuid", "asdf", "--creator-uid", "1", "--source-url", "sljdf.com"]
-        out = {'source_url': 'sljdf.com', 'bundle_uuid': 'asdf', 'uuid': '134', 'creator_uid': 1}
-        self.assertEqual(cli.parse_args(args), out)
+        expected = {'source_url': 'sljdf.com', 'bundle_uuid': 'asdf', 'uuid': '134', 'creator_uid': 1}
+        func, api_args = cli.parse_args(args)
+        self.assertEqual(api_args, expected)
```
ttung: should have a `self.assertEqual(func, "put-files")`

and same for the subsequent cases.

sampierson: func is actually a real function now. e.g. `<bound method AddedCommand.run_from_cli of <class 'hca.dss.autogenerated.put_files.PutFiles'>>`
Can you assert against those?

sampierson: Apparently you can

test/test_dss_cli.py
```diff
@@ -192,13 +192,21 @@ def test_parsing(self):
         self.assertEqual(cli.parse_args(args), out)
 
         args = ["put-files", "--creator-uid", "1", "--source-url", "sljdf.com", "134"]
-        self.assertRaises(hca.dss.error.PrintingException, cli.parse_args, args)
+        process = subprocess.Popen(["hca"] + args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
ttung: Does this actually test against the current repo?  I would imagine this would test against the installed binary.

sampierson: Indeed.  It tests against the currently installed hca script.  The first step in the Makefile for the "test" target is to install the package.  I didn't change this.  I assume there is some good reason for it cooked up by James/Andrey.   Several of the tests are written this way. 
 eg. https://github.com/HumanCellAtlas/data-store-cli/blob/567854d0d57d286a8aa0b34f84fc8035bde54a4a/test/test_dss_cli.py#L174

test/test_dss_api.py
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+import unittest, os, sys, filecmp, shutil, tempfile, uuid
```
ttung: style nit:

preferred: one import per line.
must: sort imports by name.

hca/dss/added_command.py
```diff
@@ -1,16 +1,12 @@
...
-import requests
+import argparse, json, logging, os, re, sys
+
+import jsonschema, requests, six
 from google.auth.transport.requests import Request as GoogleAuthRequest
 from google.oauth2.credentials import Credentials as OAuth2Credentials
 from google.oauth2.service_account import Credentials as ServiceAccountCredentials
 from tweak import Config
 
+import hca.dss.cli
```
ttung: why not relative import?

test/test_dss_cli.py
ttung: Well, I guess the cost to install the package isn't that significant... Just potentially quite awkward for people not in a virtualenv.

sampierson:
I thought it was previously decided that you preferred multiple imports per line, with stdlib separated from 3rd-party separated from local imports by a blank line?  I was just trying to follow that convention.  I guess I’m remembering inaccurately.  Easy to fix.

test/test_dss_cli.py
sampierson: Lol yeah this was my trigger to start using virtualenv :/

hca/dss/composite_commands/login.py
```diff
@@ -6,6 +6,7 @@
 from google_auth_oauthlib.flow import InstalledAppFlow as GoogleAuthFlow
 from tweak import Config
 
+import hca
```
ttung: Should be a relative import.

test/__init__.py
```diff
@@ -19,3 +22,23 @@ def __exit__(self, *args):
     def captured(self):
         self.buffer.seek(0, io.SEEK_SET)
         return self.buffer.read()
+
+
+def reset_tweak_changes(f):
```
ttung: I'm not sure a function decorator is superior to a with-statement context.  The with-statement context can be applied to a narrow scope of code, whereas the function decorator has to be applied to an entire function.

However, I will defer to your judgement. :)

ttung:
I think this diff stack looks fine; I would yank the staging changes out though, since they're not pertinent to the refactor even though they motivate the refactor. :)

Tests are failing though, possibly because of the use of imports that refer to the current directory.

ttung:
I would change the dss-cli test to not invoke subprocess though.  Or if so, make sure you're actually invoking the hca command in the current code tree, rather than be subject to the whims of a user's path.

sampierson:
I would like to have my cake and eat it.  I want a tool that can either be used using “with” or as a decorator.
For now this seems cleaner though.
I’ll refactor as described above if I ever find a need to scope to a narrower chunk of code.

sampierson:
Re: not invoking subprocesses.  Maybe.  But later.  Right now I'm trying to do the minimum possible refactor to get this stuff squared away so I can add the staging functionality which I have a deadline to deliver.
Invoking subprocesses is the least of the problems with tests.  Higher than that on the priority list is why something so trivial takes 5 minutes to run.
But again... later.  Fixing the existing dss tests is not my priority.

sampierson:
Staging code has been moved out.
All imports are now relative.
Tests are now green.
Please approve the PR.

PR #67, opened 2017-10-09
call render_output from upload cli command

ttung:


ttung:
I changed this to call `cls._render_output(response)` in upload.py.

PR #74, opened 2017-11-02
`hca upload file` now generates dcp-type media type param

sampierson:
As per https://docs.google.com/document/d/1TqihrgXjct9aDmTJO52_gE2WlpFysB1OkG9C8exmWTw/edit#

### DIFF
--- a/hca/upload/media_type.py
+++ b/hca/upload/media_type.py
@@ -1,0 +1,92 @@
+import magic
+import operator
+import os
+import re
+
+
+class MediaType:
+
+    RFC7320_TOKEN_CHARSET = "!#$%&'*+-.^_`|~\w"
+    DCP_METADATA_JSON_FILES = ('assay', 'project', 'sample')
+
+    @classmethod
+    def from_string(cls, media_type):
+        split_on_semicolons = media_type.split(';')
+        type_info = split_on_semicolons.pop(0)
+        type_info = re.match("(?P<top_level_type>.+)/(?P<subtype>[^+]+)(?P<suffix>\+.+)?", type_info)
+        parameters = {}
+        for parameter in split_on_semicolons:
+            param_match = re.match('(.+)=(.+)', parameter.strip())
+            parameters[param_match.group(1)] = param_match.group(2).strip('"')
+        return cls(
+            type_info.group('top_level_type'), type_info.group('subtype'),
+            suffix=type_info.group('suffix'), parameters=parameters)
+
+    @classmethod
+    def from_file(cls, file_path, dcp_type=None):
+        media_type_string = cls._media_type_from_magic(file_path)
+        media_type = cls.from_string(media_type_string)
+        if dcp_type:
+            media_type.parameters['dcp-type'] = dcp_type
+        else:
+            media_type.parameters['dcp-type'] = cls._dcp_media_type_param(media_type, os.path.basename(file_path))
+        return media_type
+
+    def __init__(self, top_level_type, subtype, suffix=None, parameters={}):
+        self.top_level_type = top_level_type
+        self.subtype = subtype
+        self.suffix = suffix
+        self.parameters = parameters
+
+    def __str__(self):
+        result = [self.main_type]
+        if self.parameters:
+            sorted_params = sorted(self.parameters.items(), key=operator.itemgetter(0))
+            for k, v in sorted_params:
+                result.append('{key}={value}'.format(key=k, value=self._properly_quoted_parameter_value(v)))
+        return '; '.join(result)
+
+    @property
+    def main_type(self):
+        maintype = "{type}/{subtype}".format(type=self.top_level_type, subtype=self.subtype)
+        if self.suffix:
+            maintype += self.suffix
+        return maintype
+
+    @staticmethod
+    def _media_type_from_magic(file_path):
+        """
+        Use libmagic to generate a media-type for the file, then correct its mistakes.
+        """
+        media_type = magic.from_file(file_path, mime=True)
+        if media_type == 'text/plain':  # libmagic doesn't recognize JSON
+            if file_path.endswith('.json'):
+                media_type = 'application/json'
+        elif media_type == 'application/x-gzip':  # deprecated
+            media_type = 'application/gzip'
+        return media_type
+
+    @staticmethod
+    def _dcp_media_type_param(media_type, filename):
+        if media_type.main_type == 'application/json':
+            (filename_without_extension, ext) = os.path.splitext(filename)
+            if filename_without_extension in MediaType.DCP_METADATA_JSON_FILES:
+                return "metadata/{filename}".format(filename=filename_without_extension)
+            else:
+                return "metadata"
+        else:
+            return "data"
+
+    @staticmethod
+    def _properly_quoted_parameter_value(value):
+        """
+        Per RFC 7321 (https://tools.ietf.org/html/rfc7231#section-3.1.1.1):
+        Parameters values don't need to be quoted if they are a "token".
+        Token characters are defined by RFC 7320 (https://tools.ietf.org/html/rfc7230#section-3.2.6).
+        Otherwise, parameters values can be a "quoted-string".
+        So we will quote values that contain characters other than the standard token characters.
+        """
+        if re.match("^[{charset}]*$".format(charset=MediaType.RFC7320_TOKEN_CHARSET), value):
+            return value
+        else:
+            return '"{}"'.format(value)
--- a/hca/upload/upload_area.py
+++ b/hca/upload/upload_area.py
@@ -1,7 +1,9 @@
+import os
 import re
 
 from .api_client import ApiClient
 from .exceptions import UploadException
+from .s3_agent import S3Agent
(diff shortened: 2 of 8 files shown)


hca/upload/upload_area.py
```diff
@@ -58,3 +62,10 @@ def forget(self):
     def list(self):
         client = ApiClient(self.urn.deployment_stage)
         return client.list_area(self.uuid)
+
+    def upload_file(self, file_path, target_filename=None, report_progress=False, dcp_type=None):
```
ttung: super minor nit: would prefer content_type/dcp_type to be the earlier argument (as opposed to `report_progress`), kind of like how s3agent does it.  it's a more important argument.

hca/upload/media_type.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class MediaType:
+
+    RFC7320_TOKEN_CHARSET = "!#$%&'*+-.^_`|~\w"
+    DCP_METADATA_JSON_FILES = ('assay', 'project', 'sample')
+
+    @classmethod
+    def from_string(cls, media_type):
+        split_on_semicolons = media_type.split(';')
```
ttung: if you have semicolons in the media type k/v pairs sections, will this break?

hca/upload/media_type.py
```diff
@@ -0,0 +1,90 @@
...
+        return maintype
+
+    @staticmethod
+    def _media_type_from_magic(file_path):
+        """
+        Use libmagic to generate a media-type for the file, then correct its mistakes.
+        """
+        media_type = magic.from_file(file_path, mime=True)
+        if media_type == 'text/plain':  # libmagic doesn't recognize JSON
+            if re.search('.json$', file_path):
```
ttung: would just do:

`file_path.endswith('.json')`

hca/upload/media_type.py
```diff
@@ -0,0 +1,90 @@
...
+        return media_type
+
+    def __init__(self, top_level_type, subtype, suffix=None, parameters={}):
+        self.top_level_type = top_level_type
+        self.subtype = subtype
+        self.suffix = suffix
+        self.parameters = parameters
+
+    def __str__(self):
+        media_type = self.main_type
```
ttung: consider:

```
result = [self.main_type]
if self.parameters:
    for k, v in self.parameters.items():
        result.append('{key}={value}'.format(key=k, value=self._properly_quoted_parameter_value(v)))
return '; '.join(result)
```

hca/upload/media_type.py
```diff
@@ -0,0 +1,90 @@
...
+        if self.parameters:
+            for k, v in self.parameters.items():
+                media_type += '; {key}={value}'.format(key=k, value=self._properly_quoted_parameter_value(v))
+        return media_type
```
ttung: suffix does not figure into here?

hca/upload/media_type.py
```diff
@@ -0,0 +1,90 @@
...
+            if re.search('.json$', file_path):
+                media_type = 'application/json'
+        elif media_type == 'application/x-gzip':  # deprecated
+            media_type = 'application/gzip'
+        return media_type
+
+    @staticmethod
+    def _dcp_media_type_param(media_type, filename):
+        if media_type.main_type == 'application/json':
+            filename_without_extension = re.search('(.+)\.[^.]', filename).group(1)
```
ttung: consider https://docs.python.org/2/library/os.path.html#os.path.splitext

test/upload/python_bindings/test_upload_file.py
```diff
@@ -28,10 +24,12 @@ def test_file_upload(self):
...
         s3.Bucket(TEST_UPLOAD_BUCKET).create()
 
-        upload.upload_file('LICENSE')
+        file_path = os.path.join(os.path.dirname(__file__), '..', '..', 'bundle', 'assay.json')
+        upload.upload_file(file_path)
 
-        obj = s3.Bucket(TEST_UPLOAD_BUCKET).Object("{}/LICENSE".format(area.uuid))
-        with open('LICENSE', 'rb') as fh:
+        obj = s3.Bucket(TEST_UPLOAD_BUCKET).Object("{}/assay.json".format(area.uuid))
+        self.assertEqual(obj.content_type, 'application/json; dcp-type="metadata/assay"')
```
ttung: 👍

test/upload/test_media_type.py
```diff
@@ -0,0 +1,57 @@
...
+        },
+        'application/json+zip': {
+            'top_level_type': 'application', 'subtype': 'json', 'suffix': '+zip',
+        },
+        'application/octet-stream+zip; dcp-type=data': {
+            'top_level_type': 'application', 'subtype': 'octet-stream',
+            'suffix': '+zip', 'parameters': {'dcp-type': 'data'}
+        },
+        'application/json; dcp-type="metadata/sample"': {
+            'top_level_type': 'application', 'subtype': 'json', 'parameters': {'dcp-type': 'metadata/sample'}
```
ttung: consider doing a test with multiple parameters.

sampierson:
Silly GitHub, stop closing my PRs automatically.

hca/upload/media_type.py
sampierson: It is valid to have multiple sets of k/v params separated by semicolons.  This works in that case.  I have added a test case for that.

For the case where a semicolon is inside a quoted string, yes this will break in that case.  Given that the use of parameters is pretty obscure anyway, I'm not sure it's worth the effort of spinning up a full blown parser which I think is what you would need to do to handle this case.  Unless you know a shortcut?

hca/upload/media_type.py
sampierson: 👍

hca/upload/media_type.py
sampierson: With `media_type = self.main_type`
`main-type` is an @property that includes the suffix if there is one

hca/upload/media_type.py
sampierson: 👍

hca/upload/media_type.py
sampierson: 👍

test/upload/test_media_type.py
sampierson: 👍

hca/upload/upload_area.py
sampierson: 👍

hca/upload/media_type.py
ttung: this syntax looks remarkably like the cookie header syntax -- pairs separated by semicolons, values can be quoted.  https://docs.python.org/2/library/cookie.html

ttung: if the cookie library doesn't work for your purposes, and you're going to assume that ';' is an uncommon value, i would suggest that you flag it when someone attempts to set a value with a ';'.

sampierson: Makes sense.  Let me look at the cookie stuff.

hca/upload/media_type.py
```diff
@@ -0,0 +1,92 @@
...
+            sorted_params = sorted(self.parameters.items(), key=operator.itemgetter(0))
+            for k, v in sorted_params:
+                result.append('{key}={value}'.format(key=k, value=self._properly_quoted_parameter_value(v)))
+        return '; '.join(result)
+
+    @property
+    def main_type(self):
+        maintype = "{type}/{subtype}".format(type=self.top_level_type, subtype=self.subtype)
+        if self.suffix:
+            maintype += self.suffix
```
ttung: I think this should be `maintype += "+" + self.suffix`

test/upload/test_media_type.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def test_string_parsing(self):
+        for media_type, attributes in self.TEST_STRINGS_AND_ATTRS.items():
+            mt = MediaType.from_string(media_type)
+            self.assertEqual(attributes['top_level_type'], mt.top_level_type)
+            self.assertEqual(attributes['subtype'], mt.subtype)
+            if 'suffix' in attributes:
+                self.assertEqual(attributes['suffix'], mt.suffix)
+            if 'parameters' in attributes:
+                self.assertEqual(attributes['parameters'], mt.parameters)
```
ttung: consider turning `mt` back into a str and compare it with the original (although this won't quite work with multiple parameters)

hca/upload/media_type.py
sampierson: I could break that out.  Currently the "+" is kept in the suffix, as it makes the regex simpler.

ttung: Ohh, I didn't realize suffix included the "+".  That is weird. :)

test/upload/test_media_type.py
sampierson: Actually that should work now.   I sort parameters before outputting now.

ttung: 👍

hca/upload/media_type.py
sampierson: Yeah it is unintuitive.  It will make it more understandable to break out the "+".

Issue #60, opened 2017-09-29
HCA CLI invocation in tests subject to whims of user's path


ttung:
Either we should be invoking the HCA CLI with a full path to the current tree or we should not be using subprocess.

kislyuk:
Fixed.

Issue #56, opened 2017-09-25
Add Search Abstraction for Paged Results


Bento007:
Add a function to retrieve all paged results from a search using post_search and get_search.

kislyuk:
Fixed.

PR #77, opened 2017-11-20
Refactor CLI client

### LINKED ISSUE #36: Support configuration/specification of DSS API endpoint for command line tools
The command line tools should support configuration/specification of the DSS API endpoint to use.
This will be important, for example, as we add deployments for dev, stage and prod.
It is also important when there are independent deployments for developer testing, etc.
The bindings currently enable an endpoint to be specified, which is great, yet the command line tools do not.
It would be convenient for a default endpoint to be configured by a user (e.g. via an env var), with the ability to specify a different endpoint as a command-line option.

kislyuk:
Sorry about the giant diff.

Refactor the library.

- Introduce a new class, hca.util.SwaggerClient, that manages a
  Swagger API connection client and dynamically generates the client
  bindings and CLI parser for any Swagger API given to it. The only
  user of this class is currently hca.dss.DSSClient.

- The class takes only the Swagger definition URL as its
  configuration, and caches the API definition in the user config
  directory after the first run. This allows dynamic configuration of
  API endpoints and other API parameters. Any higher level commands
  that should be added on top of the API bindings can be defined as
  methods in hca.util.SwaggerClient subclasses, and added to the
  `commands` attribute.

- In the new class, enable a new transport subroutine. It is a wrapper
  around Requests, as before, but now it allows explicit, idiomatic
  control over pagination, streaming, and authentication.

- The new class enables better autodoc generation. Run `make docs`;
  `open _build/html/index.html` to see the updated autodocs.

- Remove code generation machinery.

Connects to https://github.com/HumanCellAtlas/data-store/issues/468

Connects to https://github.com/HumanCellAtlas/data-store/issues/531

Fixes #36, #56, #60


hca/util/__init__.py
```diff
@@ -0,0 +1,332 @@
+"""
```
ttung: This module contains the core of the entire CLI, yet has barebones documentation.  I don't think this is in a state where other people would be able to come in and be able to work on this codebase.

hca/dss/__init__.py
```diff
@@ -1,172 +1,139 @@
...
+from ..util import SwaggerClient
+from ..util.exceptions import SwaggerAPIException
+from .. import logger
+from .upload_to_cloud import upload_to_cloud
+
+class DSSClient(SwaggerClient):
+    """
+    Client for the Data Storage Service API.
+    """
+    UPLOAD_BACKOFF_FACTOR = 1.618
```
Bento007: out of curiosity where does this magic number come from?

hca/util/__init__.py
kislyuk: Great point, sorry about that! I've added a docstring, please take a look.

hca/dss/__init__.py
kislyuk: It doesn't really matter, the exponential backoff factor for HTTP request backoffs can be any number not too far from 2.

hca/dss/__init__.py
```diff
@@ -1,172 +1,139 @@
...
+                                fh.write(chunk)
+                    logger.info("%s", "File {}: GET SUCCEEDED. Stored at {}.".format(filename, file_path))
+
+                else:
+                    logger.error("%s", "File {}: GET FAILED.".format(filename))
+                    logger.error("%s", "Response: {}".format(response.text))
+            finally:
+                response.close()
+                session.stream = False
+        return {}
```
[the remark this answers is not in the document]
kislyuk: I think it's preferable if all commands return JSON objects on their standard output by default.

hca/dss/__init__.py
```diff
@@ -1,172 +1,139 @@
...
+        bundle_uuid = str(uuid.uuid4())
+        files_to_upload, files_uploaded = [], []
+        for filename in os.listdir(src_dir):
+            full_file_name = os.path.join(src_dir, filename)
+            files_to_upload.append(open(full_file_name, "rb"))
+
+        logger.info("Uploading %i files from %s to %s", len(files_to_upload), src_dir, staging_bucket)
+        file_uuids, uploaded_keys = upload_to_cloud(files_to_upload, staging_bucket=staging_bucket, replica=replica,
+                                                    from_cloud=False)
+        for file_handle in filter(lambda x: not isinstance(x, str), files_to_upload):
```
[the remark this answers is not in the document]
kislyuk: It used to be that this could be an S3 URL prefix, but with this refactor it's no longer possible to pass that in. Thanks for the catch, I've removed the filter.

kislyuk:
@Bento007, I have added a test case for the paginator.

hca/__init__.py
```diff
@@ -1 +1,26 @@
...
+    @property
+    def config_files(self):
+        return [self.default_config_file] + tweak.Config.config_files.fget(self)
+
+    @property
+    def user_config_dir(self):
+        return os.path.join(self._user_config_home, self._name)
+
+
+_config = None
```
ttung: why not just:

`_config = HCAConfig(__name__)`

I don't think you get much from lazy-initializing it.

hca/util/exceptions.py
```diff
@@ -0,0 +1,25 @@
...
+from requests.exceptions import HTTPError
+
+class SwaggerAPIException(HTTPError):
+    """
+    Exception raised by SwaggerClient when an HTTP error is received from a Swagger API server.
+    """
+    def __init__(self, *args, **kwargs):
+        super(SwaggerAPIException, self).__init__(*args, **kwargs)
+
+        self.code = self.response.status_code
```
ttung: there's a standardized template for responses that comes as part of the json.  https://github.com/HumanCellAtlas/data-store/blob/b9471df71eeba3074a43ca50caee5c0e144f0f15/dss-api.yml#L998

hca/cli.py
```diff
@@ -2,55 +2,91 @@
...
-    def _setup_arg_parser(self):
-        self.top_parser = argparse.ArgumentParser(description=__doc__,
-                                                  formatter_class=argparse.RawDescriptionHelpFormatter)
-        top_subparsers = self.top_parser.add_subparsers(title="Commands",
-                                                        help="Use \"hca <command> -h\" to get detailed command help.\"")
+def main(args=None):
+    parser = HCAArgumentParser(description=__doc__, formatter_class=argparse.RawTextHelpFormatter)
+    parser.add_argument("--version", action="version", version="%(prog)s {version}".format(version=__version__))
+    parser.add_argument("--log-level", default=get_config().get("log_level"),
+                        help=str([logging.getLevelName(i) for i in range(0, 60, 10)]),
```
ttung: is there really a need for NOTSET?

hca/util/__init__.py
```diff
@@ -1,5 +1,84 @@
...
+
+Each user class should have a configuration subtree keyed by its name
+(such as ``APIClient`` above) under the DCP-wide config manager
+(available via ``hca.get_config()``; the static defaults for the
+config manager are stored in ``hca/default_config.json``). Within that
+subtree, the key ``swagger_url`` should point to the HTTPS URL
+containing the Swagger API definition that the client is providing an
+interface for. On first use, this API definition will be downloaded
+and saved into the user config directory (for example,
+/Users/Alice/.config/hca) with a name determined by the base64
```
ttung: is there an update strategy?

hca/util/__init__.py
ttung: The added documentation is a great step, but mostly for someone who wants to understand what this module does.  It's not really that useful for understanding _how_ it works.

If I were to come in and try to work on this, I'd probably have to drop `print(..)` statements everywhere to understand that.

hca/__init__.py
kislyuk: I do think it's beneficial to lazy-initialize it, to avoid import-time side effects that include reading a bunch of files and registering an atexit handler that will write a bunch of files.

hca/cli.py
kislyuk: Great point, removed.

hca/util/__init__.py
kislyuk: There is not yet an update strategy. I imagine an update strategy could involve re-downloading the Swagger API definition when the hca package version has changed with respect to what's recorded in the user-specific config file, or if the cached API definition file's mtime is older than a certain time cutoff. But I'd like to land that in a separate PR. Filed #79 to keep track of this.

hca/util/exceptions.py
kislyuk: Added error response fields to exception.

hca/upload/cli/list_areas_command.py
```diff
@@ -4,12 +4,15 @@
 
 
 class ListAreasCommand:
```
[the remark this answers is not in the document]
kislyuk: This should be fixed now. In the new CLI entry point dispatch convention, the return value of the call to `entry_point` is printed on stdout if it's a string or bytes, or printed as JSON if it's a mapping. The upload commands are using a class constructor call (__new__/__init__), which returns an instance of the command class, which is not something that should be printed. I special-cased this in the entry point dispatcher to ignore the return value in this case. Alternatively, we could change the Command classes to move the command handler logic from `__init__` to `__call__`, and change the entry points to be instances of the Command class instead of the classes themselves.

hca/upload/cli/list_area_command.py
```diff
@@ -2,11 +2,13 @@
...
-
+    """
+    List contents of currently selected upload area.
+    """
     @classmethod
     def add_parser(cls, staging_subparsers):
         list_area_parser = staging_subparsers.add_parser(
-            'list', description="List contents of currently selected upload area.")
+            'list', description=cls.__doc__, help=cls.__doc__)
         list_area_parser.set_defaults(func=ListAreaCommand)
```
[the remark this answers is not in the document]
kislyuk: This should be fixed now. I forgot to update all the locations in the upload source tree to change `func` to `entry_point`. I'll add a CLI test to make sure this code path is asserted on.

PR #78, opened 2017-11-30
Update GCP service account credentials; switch to env var

kislyuk:


kislyuk:
This was a red herring

PR #87, opened 2018-01-10
Retry GET requests

ttung:
1. Add a retry policy class that dictates which requests are retried and how often they are retried.
2. Add a test that uses the fake 504 mechanism in data-store to test the retry policy.

Connects to https://github.com/HumanCellAtlas/data-store/issues/485

### DIFF
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -91,19 +91,22 @@ except ImportError:
     from funcsigs import signature, Signature, Parameter
 
 import requests
+from requests.adapters import HTTPAdapter
 from requests_oauthlib import OAuth2Session
+from urllib3.util import retry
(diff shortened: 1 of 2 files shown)


hannes-ucsc:
Let me plug https://github.com/HumanCellAtlas/data-store/blob/master/dss/util/retry.py#L14 here which could also be used for this if it weren't in a different project.

ttung:
> Let me plug https://github.com/HumanCellAtlas/data-store/blob/master/dss/util/retry.py#L14 here which could also be used for this if it weren't in a different project.

time for `HumanCellAtlas/dcp-py-common`

ttung:
Updated to use urllib3 retry.

PR #88, opened 2018-01-10
Retry non-GET methods marked idempotent.

ttung:
Note that this will not work until https://github.com/HumanCellAtlas/data-store/pull/842 is deployed.

Connects to https://github.com/HumanCellAtlas/data-store/issues/485


ttung:
Don't need this.

PR #90, opened 2018-01-22
Import key modules at top level of hca package

kislyuk:
Also clean up hca namespace a bit.

### DIFF
--- a/hca/__init__.py
+++ b/hca/__init__.py
@@ -1,28 +1,4 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import os, logging
-
-from tweak import Config as _Config
-
-class HCAConfig(_Config):
-    default_config_file = os.path.join(os.path.dirname(__file__), "default_config.json")
-    @property
-    def config_files(self):
-        return [self.default_config_file] + _Config.config_files.fget(self)
-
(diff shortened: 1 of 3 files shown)


hca/__init__.py
```diff
@@ -2,13 +2,13 @@
 
 import os, logging
 
-import tweak
+from tweak import Config as _Config
```
ttung: why this change?

hca/__init__.py
```diff
@@ -24,3 +24,5 @@ def get_config():
 
 
 logger = logging.getLogger(__name__)
+
+from . import dss, upload  # noqa
```
ttung: Rather than do the `# noqa` thing, you can move the code in `__init__.py` into a module in the hca package.

hca/__init__.py
kislyuk: I'm importing it with an underscore because the Config object should not be part of this package's API.

hca/__init__.py
kislyuk: Good point, I'll try and refactor this.

hca/__init__.py
ttung: Ah!  Makes sense.

PR #91, opened 2018-01-22
Avoid crashing when writing of config file fails

kislyuk:


hca/util/__init__.py
```diff
@@ -208,19 +203,20 @@ def swagger_spec(self):
...
                 swagger_filename = base64.urlsafe_b64encode(swagger_url.encode()).decode() + ".json"
                 swagger_filename = os.path.join(self.config.user_config_dir, swagger_filename)
-            if not os.path.exists(swagger_filename):
-                try:
-                    os.makedirs(self.config.user_config_dir)
-                except OSError as e:
-                    if not (e.errno == errno.EEXIST and os.path.isdir(self.config.user_config_dir)):
-                        raise
+            if os.path.exists(swagger_filename):
+                with open(swagger_filename) as fh:
```
ttung: technically, you should wrap this in a try-catch block if your goal is to make it not crash.  file existence is not sufficient.

I would just do:

```
try:
    with open(swagger_filename) as fh:
        self.__class__._swagger_spec = json.load(fh)
except Exception:
    try:
        Path(self.config.user_config_dir).mkdir(parents=True, exist_ok=True)
        with open(swagger_filename, "wb") as fh:
            fh.write(res.content)
    except IOError as e:
        logger.error(e)
    self.__class__._swagger_spec = res.json()
```

kbergin:
Bumping on open PRs that haven't been updated in a while. Anything to do here?

PR #93, opened 2018-01-22
upload: fix list command to work for areas with many files

sampierson:
Instead of getting the file list through the Upload API, get file list directly from S3.  This means regular listings are very fast and we avoid proxied pagination.

If we are doing long listings, get detail on batches of files from the Upload API, so we don't have to give users read access to files.  Batch 100 files at a time, so the UI is fairly response (and we don't blow API gateway timeouts).

### DIFF
--- a/hca/upload/api_client.py
+++ b/hca/upload/api_client.py
@@ -14,7 +14,9 @@ class ApiClient:
         url = "{api_url_base}/area/{uuid}/files_info".format(api_url_base=self.api_url_base, uuid=area_uuid)
         response = requests.put(url, data=(json.dumps(file_list)))
         if not response.ok:
-            raise RuntimeError("GET {url} returned {status}, {content}".format(url=url,
-                                                                               status=response.status_code,
-                                                                               content=response.content))
+            raise RuntimeError(
+                "GET {url} returned {status}, {content}".format(
+                    url=url,
+                    status=response.status_code,
+                    content=response.content))
         return response.json()
--- a/hca/upload/api_client.py
+++ b/hca/upload/api_client.py
@@ -1,3 +1,5 @@
+import json
+
 import requests
 
 
@@ -8,8 +10,11 @@ class ApiClient:
     def __init__(self, deployment_stage):
         self.api_url_base = ApiClient.UPLOAD_SERVICE_API_URL_TEMPLATE.format(deployment_stage=deployment_stage)
 
-    def list_area(self, area_uuid):
-        url = "{api_url_base}/area/{uuid}".format(api_url_base=self.api_url_base, uuid=area_uuid)
-        response = requests.get(url)
-        assert response.ok
-        return response.json()['files']
+    def files_info(self, area_uuid, file_list):
+        url = "{api_url_base}/area/{uuid}/files_info".format(api_url_base=self.api_url_base, uuid=area_uuid)
+        response = requests.put(url, data=(json.dumps(file_list)))
+        if not response.ok:
+            raise RuntimeError("GET {url} returned {status}, {content}".format(url=url,
+                                                                               status=response.status_code,
+                                                                               content=response.content))
+        return response.json()
--- a/hca/upload/s3_agent.py
+++ b/hca/upload/s3_agent.py
@@ -53,6 +53,12 @@ class S3Agent:
             self.cumulative_bytes_transferred = 0
             obj.upload_fileobj(fh, **upload_fileobj_args)
 
+    def list_bucket_by_page(self, bucket_name, key_prefix):
+        paginator = self.s3.meta.client.get_paginator('list_objects')
+        for page in paginator.paginate(Bucket=bucket_name, Prefix=key_prefix, PaginationConfig={'PageSize': 100}):
+            if 'Contents' in page:
+                yield [o['Key'] for o in page['Contents']]
+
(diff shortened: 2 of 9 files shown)


hca/upload/api_client.py
```diff
@@ -8,8 +10,11 @@ class ApiClient:
...
         self.api_url_base = ApiClient.UPLOAD_SERVICE_API_URL_TEMPLATE.format(deployment_stage=deployment_stage)
 
-    def list_area(self, area_uuid):
-        url = "{api_url_base}/area/{uuid}".format(api_url_base=self.api_url_base, uuid=area_uuid)
-        response = requests.get(url)
-        assert response.ok
-        return response.json()['files']
+    def files_info(self, area_uuid, file_list):
+        url = "{api_url_base}/area/{uuid}/files_info".format(api_url_base=self.api_url_base, uuid=area_uuid)
+        response = requests.put(url, data=(json.dumps(file_list)))
```
ttung: I assume this is because GET has a lower payload size limit than PUT?  Kind of unfortunate... :(

hca/upload/api_client.py
```diff
@@ -8,8 +10,11 @@ class ApiClient:
...
+        if not response.ok:
+            raise RuntimeError("GET {url} returned {status}, {content}".format(url=url,
+                                                                               status=response.status_code,
```
ttung: recommend putting the first parameter on the next line, since that ensures you don't need to re-indent all the lines if you change the first line in the future.

hca/upload/s3_agent.py
```diff
@@ -53,6 +53,12 @@ def upload_file(self, local_path, target_bucket, target_key, content_type, repor
             self.cumulative_bytes_transferred = 0
             obj.upload_fileobj(fh, **upload_fileobj_args)
 
+    def list_bucket_by_page(self, bucket_name, key_prefix):
+        paginator = self.s3.meta.client.get_paginator('list_objects')
+        for page in paginator.paginate(Bucket=bucket_name, Prefix=key_prefix, PaginationConfig={'PageSize': 100}):
```
ttung: is there a reason that the default page size is not sufficient?

test/upload/cli/test_list_area.py
```diff
@@ -16,38 +18,49 @@
...
         with requests_mock.mock() as m:
-            mock_url = 'https://upload.test.data.humancellatlas.org/v1/area/{uuid}'.format(uuid=area.uuid)
-            m.get(mock_url, text='{"files":['
+            mock_url = 'https://upload.{stage}.data.humancellatlas.org/v1/area/{uuid}/files_info'.format(
+                stage=self.deployment_stage,
+                uuid=area.uuid)
+            m.put(mock_url, text='['
                                  '{"name":"file1.fastq.gz",'
-                                 '"content_type":"foo/bar",'
+                                 '"content_type":"binary/octet-stream dcp-type=data",'
```
ttung: I thought your format is `binary/octet-stream; dcp-type=data` (note the semi-colon)

hca/upload/api_client.py
sampierson: I was originally using GET, but I discovered that the AWS API Gateway throws away your payload for a GET, the HTTP spec says that a server response to a GET should not vary based on the payload, and as Andrey pointed out, if it does that breaks caching.  So in this case we have to abandon RESTfullness and just call it an RPC call :)

hca/upload/api_client.py
sampierson: Good idea. Will do.

hca/upload/s3_agent.py
sampierson: Yes.  Default page size is 1000.  Attempting to get file info on 1000 S3 objects from the Upload Service lambda took a considerable amount of time and was timing out at the API gateway (30s) and sometimes the Lambda timeout too (300s).  It also led to very log pauses in what should be a moderately interactive command.  Reducing to 100 keeps it within the API gateway timeout and makes the command feel more interactive.

I considered getting 1000 at a time from S3 then looping through that 100 at a time to get file info from the Upload Service but frankly I don't think that optimization is worth the extra code complexity.  99% of the work is talking to the Upload Service anyway.

test/upload/cli/test_list_area.py
sampierson: You are correct.  Will fix.

hca/upload/s3_agent.py
ttung: Cool, this makes sense, and thanks for the explanation.

test/upload/cli/test_list_area.py
```diff
@@ -16,38 +18,49 @@
...
                                  '"checksums":{"sha1":"shaaa"}}'
-                                 ']}')
+                                 ']')
 
             with CapturingIO('stdout') as stdout:
                 ListAreaCommand(Namespace(long=True))
 
         self.assertRegexpMatches(stdout.captured(), "size\s+123")
-        self.assertRegexpMatches(stdout.captured(), "Content-Type\s+foo/bar")
+        self.assertRegexpMatches(stdout.captured(), "Content-Type\s+binary/octet-stream dcp-type=data")
```
ttung: i assume this also needs the semicolon.

sampierson:
Thank you for the code review!

PR #94, opened 2018-01-23
Remove hardcoded HCA dependencies from upload CLI

sampierson:
upload: make upload_service_api_url_template configurable
Also:
- Clean up default setting.
- Actually abort when there is not current_area.
- Remove a couple of references to HCA.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,7 +2,7 @@ flake8
 moto
 coverage
 pyyaml
-requests_mock
+responses
 backports.tempfile
 mock
 -r requirements.txt
--- a/test/upload/cli/test_config.py
+++ b/test/upload/cli/test_config.py
@@ -1,0 +1,64 @@
+import os
+import sys
+import unittest
+from argparse import Namespace
+
+import responses
+import boto3
+from moto import mock_s3
+
+from ... import CapturingIO, reset_tweak_changes
+from .. import mock_current_upload_area
+
+import hca
+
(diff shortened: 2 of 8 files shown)


requirements-dev.txt
```diff
@@ -3,6 +3,7 @@ moto
 coverage
 pyyaml
 requests_mock
+responses
```
kislyuk: We have requests_mock and responses here. They both seem to be responsible for requests mocking. Are they both necessary?

test/upload/cli/test_config.py
```diff
@@ -0,0 +1,67 @@
...
+import boto3
+from moto import mock_s3
+
+from ... import CapturingIO, reset_tweak_changes
+from .. import mock_current_upload_area
+
+import hca
+
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..', '..'))  # noqa
+sys.path.insert(0, pkg_root)  # noqa
```
kislyuk: The path hack on L15-16 is intended to make `hca` imports work by adding the `dcp-cli` directory to the module search path. Clearly it's not needed here since you `import hca` on the line directly above. Also, because this test is in a sub-sub-directory of `test`, you have one `'..'` too few.

Please also check for this in other test files in this directory.

sampierson: Nice catch.  Too much copy-pasta.

requirements-dev.txt
sampierson: I removed requests_mock.

PR #97, opened 2018-03-30
Retry on read errors.

ttung:


briandoconnor:
@ttung are there any other places in the code base that the retry policy needs to be tweaked or is this it?

ttung:
> @ttung are there any other places in the code base that the retry policy needs to be tweaked or is this it?

shouldn't be.

ttung:
We may need to tweak the retry policy more as we see what kind of errors we encounter / how often we encounter them.

kislyuk:
Released in v3.4.2, please test.

PR #98, opened 2018-04-02
Retry streaming read errors in bundle download

kislyuk:
Could use some tests and also a warning in the client's streaming API autodoc, but I'm not sure how I would set up a test case, and I'd like to postpone docs until I verify that this works in the download stress tests.

### DIFF
--- a/hca/dss/__init__.py
+++ b/hca/dss/__init__.py
@@ -6,7 +6,7 @@ import uuid
 from io import open
 
 import requests
-from requests.packages.urllib3.exceptions import ProtocolError, DecodeError, ReadTimeoutError, MaxRetryError
+from requests.packages.urllib3.exceptions import ProtocolError, DecodeError, ReadTimeoutError
 
 from ..util import SwaggerClient
 from ..util.exceptions import SwaggerAPIException
--- a/hca/dss/__init__.py
+++ b/hca/dss/__init__.py
@@ -6,6 +6,7 @@ import uuid
 from io import open
 
 import requests
+from requests.packages.urllib3.exceptions import ProtocolError, DecodeError, ReadTimeoutError, MaxRetryError
 
 from ..util import SwaggerClient
 from ..util.exceptions import SwaggerAPIException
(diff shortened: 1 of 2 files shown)


hca/dss/__init__.py
```diff
@@ -37,24 +38,21 @@ def download(self, bundle_uuid, replica, version="", dest_name=""):
...
-                if response.ok:
-                    file_path = os.path.join(dest_name, filename)
+                with self.get_file.stream(uuid=file_uuid, replica=replica) as fh, open(file_path, "wb") as dest_fh:
                     logger.info("%s", "File {}: GET SUCCEEDED. Writing to disk.".format(filename))
-                    with open(file_path, "wb") as fh:
-                        for chunk in response.iter_content(chunk_size=1024*1024):
-                            if chunk:
-                                fh.write(chunk)
+                    while True:
+                        dest_fh.write(fh.raw.read(1024*1024))
```
ttung: I assume you want to terminate this loop at some point.

hca/dss/__init__.py
```diff
@@ -37,24 +38,21 @@ def download(self, bundle_uuid, replica, version="", dest_name=""):
...
                     logger.info("%s", "File {}: GET SUCCEEDED. Stored at {}.".format(filename, file_path))
-
-                else:
-                    logger.error("%s", "File {}: GET FAILED.".format(filename))
-                    logger.error("%s", "Response: {}".format(response.text))
-            finally:
```
ttung: i think you still want the `finally` clause.

hca/dss/__init__.py
kislyuk: The loop terminates when `retries.increment()` raises an exception when the retry count is exhausted.

kislyuk: Oops. Never mind... fixing.

kislyuk: Fixed. Still new at this programming thing.

hca/dss/__init__.py
```diff
@@ -37,24 +38,24 @@ def download(self, bundle_uuid, replica, version="", dest_name=""):
             file_uuid = file_["uuid"]
             filename = file_.get("name", file_uuid)
 
-            logger.info("File %s: Retrieving...", filename)
-            response = self.get_file._request(dict(uuid=file_uuid, replica=replica), stream=True)
+            logger.info("Retrieving file %s", filename)
 
+            file_path = os.path.join(dest_name, filename)
+
+            retries = self.get_session().adapters["https://"].max_retries
```
ttung: I'm not entirely sure I agree with this, but I don't know what's better.

hca/dss/__init__.py
```diff
@@ -6,6 +6,7 @@
 from io import open
 
 import requests
+from requests.packages.urllib3.exceptions import ProtocolError, DecodeError, ReadTimeoutError, MaxRetryError
```
ttung: pls remove `MaxRetryError`.

kislyuk: Done.

hca/dss/__init__.py
kislyuk: This is retrieving the retry policy that the request logic ought to apply, and applying it manually. Arguably all this machinery might be better off in the client class, but we can move it there later. Do you have an issue with the policies being the same, or...?

PR #102, opened 2018-04-03
Bump pyOpenSSL dependency on Python 2.7

kislyuk:
This avoids incompatibilities with older versions of OpenSSL and errors like this:

    AttributeError: 'module' object has no attribute 'SSL_ST_INIT'

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,8 @@ setup(
     extras_require={
         ':python_version == "2.7"': [
             'enum34 >= 1.1.6, < 2',
-            'funcsigs >= 1.0.2, < 2'
+            'funcsigs >= 1.0.2, < 2',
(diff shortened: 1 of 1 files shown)


kislyuk:
PyOpenSSL is an optional transitive requirement of requests. It always tries to import and use pyOpenSSL instead of the native Python OpenSSL bindings, but silently proceeds if that import fails.

On Python < 2.7.9 or so, SNI is not supported by the native Python OpenSSL bindings. There are other issues with it as well. That's why it tries to do that.

PyOpenSSL < 16.2.0 induces this crash when loading OpenSSL 1.1.x. So basically if pyOpenSSL is present, we need it to be up to date.

https://github.com/requests/requests/blob/master/requests/__init__.py#L93-L102

PR #105, opened 2018-04-19
Fix CLI parsers

kislyuk:
Also add more information in `hca --version`.


mweiden:
@kislyuk for some reason Travis is having trouble checking out the repo(?)

kislyuk:
Merging because this is a hotfix and the persistent server-side test failure doesn't have to do with the nature of the changes.

PR #108, opened 2018-04-23
Fix handling of enum values in JSON request bodies

hannes-ucsc:
In Swagger's body schemas, `enum` and `type` are mutually exclusive.

Blocks https://github.com/HumanCellAtlas/data-store/pull/1178


kislyuk:
LGTM, and I think we should just consider the DCP CLI tests a type of integration test for the DSS.

PR #122, opened 2018-06-02
Resolve internal references in Swagger spec

kislyuk:


hca/util/__init__.py
```diff
@@ -202,6 +203,25 @@ def __init__(self, config=None, **session_kwargs):
...
 
+    @staticmethod
+    def load_swagger_json(swagger_json, ptr_str="$ref"):
+        """
+        Load the Swagger JSON and resolve {"$ref": "#/..."} internal JSON Pointer references.
+        """
+        refs = []
+
+        def store_refs(d):
+            if len(d) == 1 and list(d)[0] == ptr_str:
```
hannes-ucsc: Using `next(iter(d))` or `ptr_str in d` for the second term is probably faster.

hca/util/__init__.py
```diff
@@ -202,6 +203,25 @@ def __init__(self, config=None, **session_kwargs):
...
+                refs.append(d)
+            return d
+
+        swagger_content = json.load(swagger_json, object_hook=store_refs)
```
hannes-ucsc: `object_hook`, neat.

hca/util/__init__.py
kislyuk: Good idea, changed to `ptr_str in d`.

PR #124, opened 2018-06-14
Fix updated files seeping into downloads of older bundles

hannes-ucsc:


hannes-ucsc:
There is no issue connected to this.

hannes-ucsc:
@kislyuk, you might want to state your preference on #127.

PR #131, opened 2018-06-20
Check we're using Python 3.6 to release

sampierson:
The SSL versions used by Python 2 on Mac OS don't use TLSv1.2.
Only do this check for release commands, so we can still run tests under other versions.
Also add httpie as a dependency, which is required to release.


sampierson:
Found a problem with this technique.

PR #135, opened 2018-07-04
Supplying the required version to dss.upload

Bento007:


Bento007:
Here is the fix for the DSS https://github.com/HumanCellAtlas/data-store/pull/1390

Bento007:
@hannes I don't it's worth the effort. Also `upload` is convenient enough.

PR #139, opened 2018-07-20
Refetch DSS API definition weekly (#79)

### LINKED ISSUE #79: Strategy to update/automatically re-fetch API definition
With #77, the API definition is fetched on-demand on first use, but there is no update strategy. This issue is about formulating, implementing, and documenting the update strategy.

calvinnhieu:
Currently, the DCP CLI fetches the DSS API definition once the first time it is run after installation and never updates after that. This change implements a strategy to refetch the DSS API definition after 7 days since the last update determined by the cached API definition's (`~/.config/hca/<base-64-encoded-swagger-url>.json`) last modified date. The refetch frequency is contestable if you guys have other ideas. 

One minor improvement we are considering is to enable content encoding on the DSS API to enable serving a compressed version of the API definition in order to minimize HCA CLI start up times. Granted, with an update policy of 1 week, this fetch would occur relatively infrequently and may go unnoticed. Further, the performance tradeoff between reducing payload size and adding a decompression step may not be significant. This change would require a small API Gateway configuration change, be applied to the entire DSS API and the compression of response payloads would be selective and invoked only by including an appropriate `Accept-Enconding` header on the API request. 

For more details:
https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-gzip-compression-decompression.html

Fixes #79

### DIFF
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -241,7 +241,7 @@ class SwaggerClient(object):
                 swagger_filename = os.path.join(self.config.user_config_dir, swagger_filename)
             is_cached = os.path.exists(swagger_filename)
             if (not is_cached) or (is_cached and
-                                 self._get_days_since_last_modified(swagger_filename) >= self._spec_valid_for_days):
+                                   self._get_days_since_last_modified(swagger_filename) >= self._spec_valid_for_days):
                 try:
                     os.makedirs(self.config.user_config_dir)
                 except OSError as e:
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -83,7 +83,7 @@ client. Subclasses can add more commands by adding them to the
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import os, datetime, types, collections, typing, json, errno, base64, argparse
+import os, types, collections, typing, json, errno, base64, argparse
 
 try:
     from inspect import signature, Signature, Parameter
@@ -95,6 +95,7 @@ from requests.adapters import HTTPAdapter
 from requests_oauthlib import OAuth2Session
 from urllib3.util import retry, timeout
 from jsonpointer import resolve_pointer
+from datetime import datetime
 
 from .. import get_config, logger
 from .compat import USING_PYTHON2
@@ -302,8 +303,8 @@ class SwaggerClient(object):
         print("Storing access credentials")
 
     def _get_days_since_last_modified(self, filename):
-        now = datetime.datetime.now()
-        last_modified = datetime.datetime.fromtimestamp(os.path.getmtime(filename))
+        now = datetime.now()
+        last_modified = datetime.fromtimestamp(os.path.getmtime(filename))
         return (now - last_modified).days
 
     def _get_oauth_token_from_service_account_credentials(self):
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -83,7 +83,7 @@ client. Subclasses can add more commands by adding them to the
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import os, types, collections, typing, json, errno, base64, argparse
+import os, datetime, types, collections, typing, json, errno, base64, argparse
 
 try:
     from inspect import signature, Signature, Parameter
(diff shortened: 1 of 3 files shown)


hca/util/__init__.py
```diff
@@ -238,7 +239,9 @@ def swagger_spec(self):
             else:
                 swagger_filename = base64.urlsafe_b64encode(swagger_url.encode()).decode() + ".json"
                 swagger_filename = os.path.join(self.config.user_config_dir, swagger_filename)
-            if not os.path.exists(swagger_filename):
+            is_cached = os.path.exists(swagger_filename)
+            if not is_cached or (is_cached and
```
kislyuk: It's unclear here what the priority is. Please parenthesize (not is_cached).

test/test_dss_swagger_spec.py
```diff
@@ -0,0 +1,88 @@
...
+                'swagger_url': self.swagger_url
+            },
+            'swagger_filename': "test_filename"
+        })
+        self.dummy_response = requests.models.Response()
+        self.dummy_response._content = b'{"swagger": ""}'
+        self.dummy_response.status_code = 200
+
+    def test_get_swagger_spec_new(self):
+        with mock.patch('os.path.exists') as mock_exists:
```
kislyuk: I think this form might be more readable:
```python
with x as y, \
    a as b, \
    c as d:
    ...
```

kislyuk: If you're already overriding the location of the filename, maybe you don't need to mock things like os.makedirs?

calvinnhieu: Using `os.makedirs` mock to verify function calls. Mocks for `os.path.exists` and `load_swagger_json` could be replaced by creating temp dirs/files, however, I would opt in favor of using mocks so as to avoid littering temp dirs/files. Plus, your suggestion for readability helps a lot.

calvinnhieu:
Verified `os.path.getmtime` works for both read-only (444) and unreadable (000) files.

Addressing broken test now

ttung:
Can this be done with a conditional GET ([if-none-match](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match))?

I would also consider adding a cmd-line option to force a refresh of the spec.

You could make the refresh even less obtrusive by [daemonizing](https://pypi.org/project/python-daemon/) the process, and then refresh the spec after you return control to the shell.  This may not work for Windows, though.

Last, the cache for the swagger spec is not race-condition safe.  Either we should save it with some checksum, or we should save it to a tempfile, and then os.rename it to its final location (os.rename is atomic while file writes are not).

calvinnhieu:
A conditional GET would work, however, since (from my understanding) the CLI is stateless between commands, it would invoke an HTTP request (to compare ETags) for each command which I imagine is more overhead than checking a file's `mtime` (as these changes propose).

Daemonizing could resolve this performance issue and would be a requirement to using a conditional GET. Will briefly look into running daemons on Windows. Another point to consider re: asynchronously updating the cache is, with the way the update policy is written now (update cache on `swagger_spec` access), that the first command run in a batch of commands will most likely run with a stale version of the spec. To avoid this, we would effectively update the cache synchronously, negating the benefits of using a daemon. @kislyuk thoughts?

Re: CLI command to force refresh spec + use race-condition safe operations, sounds great! I'll add those today.

ttung:
> A conditional GET would work, however, since (from my understanding) the CLI is stateless between commands, it would invoke an HTTP request (to compare ETags) for each command which I imagine is more overhead than checking a file's mtime (as these changes propose).

If you save the ETag of the cached entry, you can present it to the server.  It's probably a relatively minor optimization, but if more tooling is built like this, it would be a good example to set.

> Daemonizing

I suspect UX improvement is probably not worth the complexity, but just throwing it out there. :)

calvinnhieu:
Yeah totally, appreciate the insight @ttung. What do you mean by "present it to the server"? Wouldn't that still require an HTTP request?

ttung:
> What do you mean by "present it to the server"? Wouldn't that still require an HTTP request?

yes, but we would get a smaller response if there's no change.

PR #140, opened 2018-07-23
Write/update DSS API definition atomically

calvinnhieu:
As per @ttung's suggestion (GH-139), these changes update the cached version of the DSS API definition atomically in order to avoid a race condition. This is achieved by writing the fetched file to a temp file and renamed (`os.rename`) to the desired filename, where [the latter operation is atomic](https://stackoverflow.com/questions/2333872/atomic-writing-to-file-with-python).

07/25/18 update:
These changes also handle `os.rename` [Windows specific behavior](https://docs.python.org/2/library/os.html#os.rename) by deleting the destination file if it exists before renaming. This process is attempted multiple times (5) in order to handle a potential race condition in which the destination file is recreated in between delete and rename operations. These changes also introduce `hca.util.fs_helper.py`, a static library of functions for interacting with the local filesystem.

Connected to #79

### DIFF
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -95,12 +95,13 @@ from requests.adapters import HTTPAdapter
 from requests_oauthlib import OAuth2Session
 from urllib3.util import retry, timeout
 from jsonpointer import resolve_pointer
-from datetime import datetime
 
 from .. import get_config, logger
 from .compat import USING_PYTHON2
 from .exceptions import SwaggerAPIException, SwaggerClientInternalError
 from ._docs import _pagination_docstring, _streaming_docstring, _md2rst
+from .fs_helper import FSHelper as fs
+
 
 class RetryPolicy(retry.Retry):
     def __init__(self, retry_after_status_codes={301}, **kwargs):
@@ -163,6 +164,7 @@ class _ClientMethodFactory(object):
         self._context_manager_response.close()
         self._context_manager_response = None
 
+
 class _PaginatingClientMethodFactory(_ClientMethodFactory):
     def iterate(self, **kwargs):
         page = None
@@ -170,6 +172,7 @@ class _PaginatingClientMethodFactory(_ClientMethodFactory):
             page = self._request(kwargs, url=page.links["next"]["url"] if page else None)
             for result in page.json()["results"]:
                     yield result
+
 
 class SwaggerClient(object):
     scheme = "https"
@@ -240,9 +243,9 @@ class SwaggerClient(object):
             else:
                 swagger_filename = base64.urlsafe_b64encode(swagger_url.encode()).decode() + ".json"
                 swagger_filename = os.path.join(self.config.user_config_dir, swagger_filename)
-            is_cached = os.path.exists(swagger_filename)
-            if (not is_cached) or (is_cached and
-                                   self._get_days_since_last_modified(swagger_filename) >= self._spec_valid_for_days):
+            if (("swagger_filename" not in self.config) and
+                ((not os.path.exists(swagger_filename)) or
+                 (fs.get_days_since_last_modified(swagger_filename) >= self._spec_valid_for_days))):
                 try:
                     os.makedirs(self.config.user_config_dir)
                 except OSError as e:
@@ -251,8 +254,7 @@ class SwaggerClient(object):
                 res = self.get_session().get(swagger_url)
                 res.raise_for_status()
                 assert "swagger" in res.json()
-                with open(swagger_filename, "wb") as fh:
-                    fh.write(res.content)
+                fs.atomic_write(os.path.basename(swagger_filename), self.config.user_config_dir, res.content)
             with open(swagger_filename) as fh:
                 self.__class__._swagger_spec = self.load_swagger_json(fh)
         return self._swagger_spec
@@ -303,11 +305,6 @@ class SwaggerClient(object):
                                         expires_at="-1",
                                         token_type="Bearer")
         print("Storing access credentials")
-
-    def _get_days_since_last_modified(self, filename):
-        now = datetime.now()
-        last_modified = datetime.fromtimestamp(os.path.getmtime(filename))
-        return (now - last_modified).days
 
     def _get_oauth_token_from_service_account_credentials(self):
         scopes = ["https://www.googleapis.com/auth/userinfo.email"]
--- a/hca/util/fs_helper.py
+++ b/hca/util/fs_helper.py
@@ -1,0 +1,79 @@
+import os, tempfile, errno, warnings
+from datetime import datetime
+
+
+class FSHelper(object):
+    """
+    Static helper class providing convenience methods for reading and writing files.
+    """
+
+    @staticmethod
+    def atomic_write(filename, shared_dir, content):
+        """
+        Atomically writes content a to file.
+
+        Writes to a temp file and renames file (atomic operation) to desired filename.
+        To ensure atomicity, the src and dest dirs must be the same.
+
+        :param filename: File name only (e.g. example.txt)
+        :param shared_dir: Shared directory of temp and dest file
+        :param content: Content to write to file
+        """
+        temp_filename = None
+        dest_filename = os.path.join(shared_dir, filename)
+        try:
+            with tempfile.NamedTemporaryFile('wb', dir=shared_dir, delete=False) as temp:
+                temp_filename = temp.name
+                temp.write(content)
+            FSHelper._atomic_rename(temp_filename, dest_filename)
+        finally:
+            if temp_filename and os.path.isfile(temp_filename):
+                try:
+                    os.remove(temp_filename)
+                except EnvironmentError:
+                    warnings.warn("\nFailed to clean up temporary file {}".format(temp_filename))
+
+    @staticmethod
+    def _atomic_rename(src, dest, max_retries=5):
+        """
+        Atomically renames a file.
+        Support for Unix and Windows platforms.
+
+        :param src: Source filename
+        :param dest: Destination filename
+        :param max_retries: Number of rename operation retries
+        """
+        for _ in range(max_retries):
+            try:
+                os.rename(src, dest)
+                return
+            except OSError as e:
+                FSHelper._handle_rename_error(e, src, dest)
+        warnings.warn("\nFailed to rename {} to {} after {} tries".format(src, dest, max_retries))
+
+    @staticmethod
+    def _handle_rename_error(err, src, dest):
+        """
+        Handles Windows specific rename error: https://docs.python.org/2/library/os.html#os.rename
+
+        :param err: OSError
+        :param src: Rename source filename
+        :param dest: Rename destination filename
+        """
+        if err.errno == errno.EEXIST:
+            try:
+                os.remove(dest)
+            except EnvironmentError:
+                warnings.warn("\nFailed to delete {}".format(dest))
+        else:
+            raise Exception("\nFailed to rename {} to {}".format(src, dest), err)
+
+    @staticmethod
+    def get_days_since_last_modified(filename):
+        """
+        :param filename: Absolute file path
+        :return: Number of days since filename's last modified time
+        """
+        now = datetime.now()
+        last_modified = datetime.fromtimestamp(os.path.getmtime(filename))
+        return (now - last_modified).days
--- a/test/test_dss_swagger_spec.py
+++ b/test/test_dss_swagger_spec.py
@@ -17,51 +17,54 @@ else:
 
 class TestDssSwaggerSpec(unittest.TestCase):
     client = hca.dss.DSSClient()
-    swagger_url = ""
     dummy_response = None
+    swagger_url = "https://dss.data.humancellatlas.org/v1/swagger.json"
+    swagger_filename = "test_swagger_filename"
     open_fn_name = "__builtin__.open" if USING_PYTHON2 else "builtins.open"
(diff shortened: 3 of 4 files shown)


hca/util/__init__.py
```diff
@@ -233,6 +233,7 @@ def store_refs(d):
     def swagger_spec(self):
         if not self._swagger_spec:
             swagger_url = self.config[self.__class__.__name__].swagger_url
+            temp_swagger_filename = os.path.join(self.config.user_config_dir, "temp_swagger.json")
```
ttung: I would use `with tempfile.NamedTemporaryFile`

calvinnhieu: Cool. Thanks!

kislyuk: Oof you should then use it with the option to use the user config dir as the tempdir. Otherwise the rename is not atomic either (since it could actually be a copy).

calvinnhieu:
update: `_atomic_write_to_file` using `NamedTemporaryFile` to write temp file and rename (atomic operation) to desired location

hca/util/__init__.py
```diff
@@ -302,6 +301,17 @@ def login(self, access_token=""):
...
+    def _atomic_write_to_file(self, filename, dir, content):
+        """
+        Atomically writes content a to file.
+
+        Creates a temp file and renames (atomic) file to desired filename.
+        To ensure atomicity, the src and dest dirs must be the same.
+        """
+        with tempfile.NamedTemporaryFile('wb', dir=dir, delete=False) as tmp:
+            tmp.write(content)
+            os.rename(tmp.name, filename)
```
ttung: I believe that on Windows, it's not kosher to rename a file that's open.  Consider creating the tempfile with `delete=False`, close it, and then renaming it.  Then wrap the entire try-finally with something that cleans up the tempfile in case the rename fails.

ttung: Also note from https://docs.python.org/2/library/os.html#os.rename:

> On Windows, if dst already exists, OSError will be raised even if it is a file

Here's some possible strategies for defeating that: https://bugs.python.org/issue8828

hca/util/__init__.py
```diff
@@ -302,6 +303,47 @@ def login(self, access_token=""):
...
+        temp_filename = ""
+        dest_filename = os.path.join(dir, file)
+        writing = True
+        max_retries = 5
+        retry_count = 0
+        try:
+            with tempfile.NamedTemporaryFile('wb', dir=dir, delete=False) as temp:
+                temp_filename = temp.name
+                temp.write(content)
+            while writing and retry_count < max_retries:
```
ttung: suggestion:

```
for _ in range(max_retries):
    try:
        os.rename(temp_filename, dest_filename)
        break
    except OSError as ex:
        if ex.errno == errno.EEXIST:
        # handle Windows: https://docs.python.org/2/library/os.html#os.rename
            os.remove(dest_filename)
            continue
        raise
else:
    raise Exception("Could not rename {} to {}".format(temp_filename, dest_filename))
```

you should be able to get rid of `writing` and `retry_count`

hca/util/__init__.py
```diff
@@ -302,6 +303,47 @@ def login(self, access_token=""):
...
+        Atomically writes content a to file.
+
+        Creates a temp file and renames file (atomic operation) to desired filename.
+        To ensure atomicity, the src and dest dirs must be the same.
+
+        :param file: File name only (e.g. swagger.json)
+        :param dir: Shared directory of temp and dest file
+        :param content: Content to write to file
+        """
+        temp_filename = ""
```
ttung: do you actually need this?

hca/util/__init__.py
```diff
@@ -251,8 +251,9 @@ def swagger_spec(self):
                 res = self.get_session().get(swagger_url)
                 res.raise_for_status()
                 assert "swagger" in res.json()
-                with open(swagger_filename, "wb") as fh:
-                    fh.write(res.content)
+                self._atomic_write_to_file(self._get_path_leaf(swagger_filename),
```
ttung: os.path.basename doesn't work here?

hca/util/__init__.py
calvinnhieu: Right. Python's a little rusty, thanks!

hca/util/__init__.py
calvinnhieu: Not necessarily, but it clears an IDE warning (and possibly linter?)

hca/util/__init__.py
calvinnhieu: Oh, it would - I was covering an edge case that, actually, we don't need to worry about: https://stackoverflow.com/questions/8384737/extract-file-name-from-path-no-matter-what-the-os-path-format
-> Will replace `ntpath.basename` with `os.path.basename`

However, `_get_path_leaf` handles the trailing slash case (e.g. `path/to/file/`).

hca/util/fs_helper.py
```diff
@@ -0,0 +1,80 @@
...
+        :param content: Content to write to file
+        """
+        dest_filename = os.path.join(shared_dir, filename)
+        try:
+            with tempfile.NamedTemporaryFile('wb', dir=shared_dir, delete=False) as temp:
+                temp_filename = temp.name
+                temp.write(content)
+            FSHelper._atomic_rename(temp_filename, dest_filename)
+        finally:
+            if temp_filename and os.path.isfile(temp_filename):
```
ttung: if NamedTemporaryFile throws as exception, then temp_filename will not be defined.  one possible strategy is:

```
if 'temp_filename' in locals() and os.path.isfile(temp_filename):
```

another strategy is to define temp_filename before the loop.

hca/util/fs_helper.py
```diff
@@ -0,0 +1,80 @@
...
+                os.remove(temp_filename)
```
ttung: you should wrap os.remove in a try-except EnvironmentError clause.

hca/util/fs_helper.py
```diff
@@ -0,0 +1,80 @@
...
+
+    @staticmethod
+    def _atomic_rename(src, dest):
```
ttung: suggestion:

`def _atomic_rename(src, dest, max_retries=5):`

hca/util/fs_helper.py
```diff
@@ -0,0 +1,80 @@
...
+        max_retries = 5
+        for _ in range(max_retries):
+            try:
+                os.rename(src, dest)
+                return
+            except OSError as e:
+                FSHelper._handle_rename_error(e, src, dest)
+
+    @staticmethod
+    def _handle_rename_error(err, src, dest):
```
ttung: I think you could just inline this method.  It's not particularly large and it's not called elsewhere.

hca/util/fs_helper.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+        Handles Windows specific rename error: https://docs.python.org/2/library/os.html#os.rename
+
+        :param err: OSError
+        :param src: Rename source filename
+        :param dest: Rename destination filename
+        """
+        if err.errno == errno.EEXIST:
+            os.remove(dest)
```
ttung: os.remove(..) should be wrapped in try-except EnvironmentError

hca/util/fs_helper.py
calvinnhieu: I like it

hca/util/fs_helper.py
calvinnhieu: Yea I agree, however, it's pulled out now since when it was inline in `_atomic_rename`, it violated our linter's Code Complexity limit (7 > 6)

hca/util/fs_helper.py
```diff
@@ -0,0 +1,90 @@
...
+            with tempfile.NamedTemporaryFile('wb', dir=shared_dir, delete=False) as temp:
+                temp_filename = temp.name
+                temp.write(content)
+            FSHelper._atomic_rename(temp_filename, dest_filename)
+        finally:
+            if temp_filename and os.path.isfile(temp_filename):
+                try:
+                    os.remove(temp_filename)
+                except EnvironmentError:
+                    print("Failed to clean up temporary file {}".format(temp_filename))
```
ttung: https://docs.python.org/2/library/warnings.html

hca/util/fs_helper.py
```diff
@@ -0,0 +1,90 @@
...
+                    raise
```
ttung: I don't think this is a case where we should fatal.  I would just warn and move on.

hca/util/fs_helper.py
```diff
@@ -0,0 +1,90 @@
...
+
+        :param err: OSError
+        :param src: Rename source filename
+        :param dest: Rename destination filename
+        """
+        if err.errno == errno.EEXIST:
+            try:
+                os.remove(dest)
+            except EnvironmentError:
+                print("Failed to delete {}".format(dest))
```
ttung: https://docs.python.org/2/library/warnings.html

hca/util/fs_helper.py
```diff
@@ -0,0 +1,90 @@
...
+        else:
+            print("Failed to rename {} to {}".format(src, dest))
```
ttung: `raise Exception("Failed to rename {} to {}".format(src, dest), err)`

hca/util/fs_helper.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        :param filename: Absolute file path
+        :return: Number of days since filename's last modified time
+        """
+        now = datetime.now()
+        last_modified = datetime.fromtimestamp(os.path.getmtime(filename))
+        return (now - last_modified).days
+
+    @staticmethod
+    def get_path_leaf(path):
```
ttung: this is still unnecessary as we will never pass in a path with a trailing backslash.

Makefile
```diff
@@ -3,7 +3,7 @@ SHELL=/bin/bash
 test: lint install
 	# https://github.com/HumanCellAtlas/dcp-cli/issues/127
 	coverage run --source=hca -m unittest discover -v -t . -s test/upload
-	coverage run --source=hca -m unittest discover -v -t . -s test -p test_dss_*.py
```
ttung: can you make this a separate PR and land this ASAP?  Seems like a pretty big deal.

hca/util/__init__.py
```diff
@@ -232,17 +235,18 @@ def store_refs(d):
     @property
     def swagger_spec(self):
         if not self._swagger_spec:
+            use_local_config = False
             swagger_url = self.config[self.__class__.__name__].swagger_url
             if "swagger_filename" in self.config:
                 swagger_filename = self.config.swagger_filename
                 if not swagger_filename.startswith("/"):
+                    use_local_config = True
```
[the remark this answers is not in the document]
calvinnhieu: Right. Good catch

hca/util/__init__.py
```diff
@@ -232,17 +235,18 @@ def store_refs(d):
...
                     swagger_filename = os.path.join(os.path.dirname(__file__), swagger_filename)
             else:
                 swagger_filename = base64.urlsafe_b64encode(swagger_url.encode()).decode() + ".json"
                 swagger_filename = os.path.join(self.config.user_config_dir, swagger_filename)
-            is_cached = os.path.exists(swagger_filename)
-            if (not is_cached) or (is_cached and
-                                   self._get_days_since_last_modified(swagger_filename) >= self._spec_valid_for_days):
+            if ((not use_local_config) and ((not os.path.exists(swagger_filename)) or
```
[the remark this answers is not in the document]
calvinnhieu: Right

hca/util/fs_helper.py
calvinnhieu: Thanks!

hca/util/fs_helper.py
calvinnhieu: Makes sense

Makefile
```diff
@@ -4,6 +4,7 @@ test: lint install
 	# https://github.com/HumanCellAtlas/dcp-cli/issues/127
 	coverage run --source=hca -m unittest discover -v -t . -s test/upload
 	coverage run --source=hca -m unittest discover -v -t . -s test -p test_dss_*.py
+	coverage run --source=hca -m unittest discover -v -t . -s test -p test_fs_helper.py
```
ttung: remove this change, please.

hca/util/__init__.py
```diff
@@ -240,9 +243,8 @@ def swagger_spec(self):
             else:
                 swagger_filename = base64.urlsafe_b64encode(swagger_url.encode()).decode() + ".json"
                 swagger_filename = os.path.join(self.config.user_config_dir, swagger_filename)
-            is_cached = os.path.exists(swagger_filename)
-            if (not is_cached) or (is_cached and
-                                   self._get_days_since_last_modified(swagger_filename) >= self._spec_valid_for_days):
+            if (("swagger_filename" not in self.config) and ((not os.path.exists(swagger_filename)) or
```
ttung: do you want the auto-update to happen if someone specifies `swagger_filename` in the config?

calvinnhieu: No, if someone specifies `swagger_filename` in the config, we assume that they are using a local copy of the Swagger file.  We wouldn't want to overwrite their local copy and only want to update to the latest Swagger file in the user config directory `~/hca/.config`.

Makefile
calvinnhieu: How come?

calvinnhieu: Oh got it. I'll merge https://github.com/HumanCellAtlas/dcp-cli/pull/143 first

Issue #142, opened 2018-07-27
DSSClient spontaneaously reverts to prod, even when configured with different host


hannes-ucsc:
The [prescribed](https://github.com/HumanCellAtlas/dcp-cli/blame/17e1b202f81e4fa1a0306e417ca528bbd7d150c4/README.rst#L54) method to switch the DSSClient to a different host is 

```
client = hca.dss.DSSClient()
client.host = "https://dss.example.com/v1"
```

There are two problems with this approach:

For one the swagger def is downloaded (or loaded from disk) in the constructor, defaulting to the HCA prod instance. There is no guarantee that the `dss.example.com` endpoint has a compatible API.

More puzzling is that the client seems to sporadically revert to using the prod instance, even when configured to a different host as outlined above. This must be a race condition. I do have multiple threads sharing a client.

If I instantiate the client in the following way, the problem goes away (the client does not revert to prod):

```
 config = HCAConfig("hca")
 config['DSSClient'].swagger_url = 'https://dss.example.com/v1/swagger.json'
 self.dss_client = DSSClient(config=config)
```

This is all I got, I don't have time to dig deeper right now.

hannes-ucsc:
My gut feeling is that this is related to the shared state in the DSSClient class attributes, either the urlib session or the swagger spec.

calvinnhieu:
Hey @hannes-ucsc,

I'd need to know more details about where you're setting `client.host` and when you might be creating new instances of DSSClient to be certain, but it seems like what you think is happening here is correct.

For one, the _prescribed_ method for updating the host is susceptible to this issue (and should be updated/clarified in the docs) because it does not interact with the client's config object (as you've discovered), which is loaded from and cached at `~/.config/hca/config.json` using [tweak](https://pypi.org/project/tweak/). There are two ways to update the host and have it persist:
1) The method you described (programmatically)
2) Configure the `swagger_url` field in `~/.config/hca/config.json`

What I suspect is happening in your case is that a DSSClient instance for which you set `client.host = <host>` will behave as you expect, however, any instances for which you do not specify this, will read the default `swagger_url` from `dcp-cli/hca/default_config.json` which points to prod. In either case, since the client's config object wasn't modified, at some point, `tweak` will write the instance's config object to the cache at `~/.config/hca/config.json`, most likely pointing to prod.

[...]

PR #144, opened 2018-07-30
Update docs to correct host configuration steps (#142) (#145)

calvinnhieu:
The current steps for configuring a DSS API host for local development do not work for more complex use cases such as creating multiple `DSSClient` instances with different API hosts. These changes simply update the current docs to provide a correct method. This is due to a number of reasons which speak to the need to refactor/rethink how configuration is implemented in the CLI. A separate issue should (and will) be created to outline this work. Briefly, a couple issues are:

- `_swagger_spec` is a class variable on `SwaggerClient` forcing all instances to share the same host
- CLI configuration (includes `swagger_url` is written and loaded from cache implicitly, providing the user little visibility/control.

Connected to (#142) (#145)


calvinnhieu:
@kislyuk Great! Done

Issue #149, opened 2018-08-01
dcp cli should use the constants defined in dcplib


ttung:
defined: https://github.com/HumanCellAtlas/dcplib/blob/1eb7032d01f6ac36b55c1bf3096d325978b026f2/dcplib/s3_multipart.py

currently should replace constants in https://github.com/HumanCellAtlas/dcp-cli/blob/aa5c7b6723ca01b9c9a7c723363966de9c31241b/hca/dss/upload_to_cloud.py and https://github.com/HumanCellAtlas/dcp-cli/blob/ed13906c2fa233184bd05b963f3ede2c200a9666/hca/upload/s3_agent.py

ttung:
@mweiden can you make sure this gets addressed?

mweiden:
@ttung sure

sampierson:
I have a PR open that addresses the Upload part of this
https://github.com/HumanCellAtlas/dcp-cli/pull/147

PR #150, opened 2018-08-01
Use s3 multipart constants as defined in dcplib

mweiden:
Fixes https://github.com/HumanCellAtlas/dcp-cli/issues/149
Blocked by https://github.com/HumanCellAtlas/dcplib/pull/2

### DIFF
--- a/hca/dss/upload_to_cloud.py
+++ b/hca/dss/upload_to_cloud.py
@@ -11,7 +11,8 @@ import uuid
 import boto3
 from boto3.s3.transfer import TransferConfig
 
-from dcplib.checksumming_io import ChecksummingBufferedReader, S3Etag
+from dcplib import s3_multipart
+from dcplib.checksumming_io import ChecksummingBufferedReader
 
 
 def encode_tags(tags):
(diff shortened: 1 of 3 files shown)


hca/dss/upload_to_cloud.py
```diff
@@ -27,7 +28,7 @@ def _mime_type(filename):
     return "application/octet-stream"
 
 
-def _copy_from_s3(path, s3, tx_cfg):
```
mweiden: `tx_cfg` isn't used anywhere?

mweiden:
@ttung @sampierson will take this out of WIP when https://github.com/HumanCellAtlas/dcp-cli/pull/150 is merged and `dcplib-1.3.2` is published

hca/dss/upload_to_cloud.py
```diff
@@ -27,7 +28,7 @@ def _mime_type(filename):
...
+def _copy_from_s3(path, s3):
```
ttung: kind of confused why this method has this name, but probably out of the scope of this diff.

PR #155, opened 2018-08-09
upload cli parallelization 5x improvement

parthshahva:
Results of tests:
1) 5x improvement
2) Speed now on par with aws sync

Link to test runs:
https://docs.google.com/spreadsheets/d/1KoHmrTLZrrFTtEw6_HSjFFYjNSFnngbpBLvqrNfnGzM/edit?usp=sharing


hca/upload/cli/upload_command.py
```diff
@@ -37,20 +37,12 @@ def add_parser(cls, upload_subparsers):
...
-        if report_progress:
-            print("Uploading %s to upload area %s..." % (os.path.basename(file_path), current_area_uuid))
-        upload_file(file_path, target_filename, use_transfer_acceleration=use_transfer_acceleration,
-                    report_progress=report_progress, dcp_type="data")
-        if report_progress:
-            print("\n")
+        area = UploadArea(uuid=UploadConfig().current_area)
+        area.upload_files(args.file_paths,
+                          target_filename=args.target_filename,
+                          use_transfer_acceleration=(not args.no_transfer_acceleration),
```
MDunitz: this might be outside the scope of the ticket but why are we expecting `no_transfer_acceleration` as a command line arg and then negating it when we pass it in instead of just using `transfer_acceleration` or `use_transfer_acceleration`

MDunitz: thanks!

hca/upload/upload_area.py
```diff
@@ -98,10 +100,45 @@ def list(self, detail=False):
...
+        """
+        A function that takes in a list of file paths and other optional args for parallel file upload
+        """
+        self._setup_s3_agent_for_file_upload(file_paths=file_paths, use_transfer_acceleration=use_transfer_acceleration)
+        pool = ThreadPool()
+        if report_progress:
+            print("\nStarting upload of %s files to upload area %s" % (len(file_paths), self.uuid))
+        for file_path in file_paths:
+            pool.add_task(self._upload_file, file_path,
+                          target_filename=target_filename,
```
mckinsel: Does this mean that all the `target_filename`s have to be the same for every uploaded file?

sampierson: That's a good point.  Individual filename naming doesn't with with either folder upload or parallelization.  We need to be able to turn off parallelization and prohibit folder upload if the filename option is provided.

sampierson: Marcus it sounds like this is a feature you are still using in anger?

mckinsel: It is, though there are easy workarounds. I think I'm probably unusual in that I'm using the hca python library rather than the cli tool. The latter does have some safeguards that stop you from uploading multiple files and specifying a target name.

PR #177, opened 2018-08-29
Enable programmatic SwaggerClient host configuration (#142)

### LINKED ISSUE #142: DSSClient spontaneaously reverts to prod, even when configured with different host

calvinnhieu:
Resolves host configuration issues when creating multiple SwaggerClient instances across multiple threads by:
- refactoring `_swagger_spec` field from a class var to an instance var
- enabling host configuration during client initialization

Specifically, host configuration is done via the `swagger_url` parameter in the client constructor. This is because the Swagger definition, which is served by `swagger_url`, provides information about the API and is the source of truth for the API host.

Fixes #142

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,11 +39,16 @@ the hca package by running `make install` in the repository root directory.
 
 To use the command line interface with a local or test DSS, first run ``hca`` (or ``scripts/hca`` if you want to use the
 package in place from the repository root directory). This will create the file ``~/.config/hca/config.json``, which you
-can modify to update the value of `DSSClient.swagger_url` to point to the URL of the Swagger definition served by your
-DSS deployment. Lastly, the CLI enforces HTTPS connection to the DSS API. If you are connecting to a local DSS, make this
-change in ``dcp-cli/hca/util/__init__.py`` in the `SwaggerClient` object::
+can modify to update the value of ``DSSClient.swagger_url`` to point to the URL of the Swagger definition served by your
+DSS deployment. Lastly, the CLI enforces HTTPS connection to the DSS API. If you are connecting to a local DSS, make
+this change in ``dcp-cli/hca/util/__init__.py`` in the ``SwaggerClient`` object::
 
     scheme = "http"
+
+To use the Python interface with a local or test DSS, pass the URL of the Swagger definition to the ``DSSClient``
+constructor via the ``swagger_url`` parameter::
+
+    client = DSSClient(swagger_url="https://dss.example.com/v1/swagger.json")
 
 You can also layer a minimal config file on top of the default ``config.json`` using the ``HCA_CONFIG_FILE`` environment
 variable, for example::
@@ -51,16 +56,6 @@ variable, for example::
     export SWAGGER_URL="https://dss.staging.data.humancellatlas.org/v1/swagger.json"
     jq -n .DSSClient.swagger_url=env.SWAGGER_URL > ~/.config/hca/config.staging.json
     export HCA_CONFIG_FILE=~/.config/hca/config.staging.json
-
-To use the Python interface with a local or test DSS, configure an HCAConfig object with a ``swagger_url`` field, and
-pass it as a parameter to the API client's constructor:
-
-.. code-block:: python
-
-    config = HCAConfig()
-    config['DSSClient'].swagger_url = "https://dss.example.com/v1/swagger.json"
-    client = DSSClient(config=config)
-    res = client.post_search(...)
 
 Testing
 -------
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -180,7 +180,6 @@ class SwaggerClient(object):
     retry_policy = RetryPolicy(read=10, status=10, status_forcelist=frozenset({500, 502, 503, 504}))
     _authenticated_session = None
     _session = None
-    _swagger_spec = None
     _spec_valid_for_days = 7
     _type_map = {
         "string": str,
@@ -196,9 +195,11 @@ class SwaggerClient(object):
     #
     timeout_policy = timeout.Timeout(connect=20, read=40)
 
-    def __init__(self, config=None, **session_kwargs):
+    def __init__(self, config=None, swagger_url=None, **session_kwargs):
         self.config = config or get_config()
+        self.swagger_url = swagger_url or self.config[self.__class__.__name__].swagger_url
         self._session_kwargs = session_kwargs
+        self._swagger_spec = None
 
         if USING_PYTHON2:
             self.__doc__ = _md2rst(self.swagger_spec["info"]["description"])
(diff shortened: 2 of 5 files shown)


README.rst
```diff
@@ -39,29 +39,25 @@ the hca package by running `make install` in the repository root directory.
...
+DSS deployment. Lastly, the CLI enforces HTTPS connection to the DSS API. If you are connecting to a local DSS, make
+this change in ``dcp-cli/hca/util/__init__.py`` in the ``SwaggerClient`` object::
 
     scheme = "http"
 
+To use the Python interface with a local or test DSS, pass the URL of the Swagger definition to the ``DSSClient``
+constructor via the ``swagger_url`` parameter::
+
+    client = DSSClient(swagger_url="https://dss.example.com/v1/swagger.json")
+    res = client.post_search(...)
```
mweiden: nit: I'd just leave out the call to `post_search`. It just demonstrates unrelated usage of the API, no?

README.rst
```diff
@@ -39,29 +39,25 @@ the hca package by running `make install` in the repository root directory.
...
+
 You can also layer a minimal config file on top of the default ``config.json`` using the ``HCA_CONFIG_FILE`` environment
 variable, for example::
 
     export SWAGGER_URL="https://dss.staging.data.humancellatlas.org/v1/swagger.json"
     jq -n .DSSClient.swagger_url=env.SWAGGER_URL > ~/.config/hca/config.staging.json
     export HCA_CONFIG_FILE=~/.config/hca/config.staging.json
 
-To use the Python interface with a local or test DSS, configure an HCAConfig object with a ``swagger_url`` field, and
```
mweiden: glad you moved this closer to the other discussion of the swagger endpoint 👍

hca/util/__init__.py
```diff
@@ -180,7 +180,6 @@ class SwaggerClient(object):
     retry_policy = RetryPolicy(read=10, status=10, status_forcelist=frozenset({500, 502, 503, 504}))
     _authenticated_session = None
     _session = None
-    _swagger_spec = None
```
mweiden: good catch

hca/util/__init__.py
```diff
@@ -196,9 +195,11 @@ class SwaggerClient(object):
     #
     timeout_policy = timeout.Timeout(connect=20, read=40)
 
-    def __init__(self, config=None, **session_kwargs):
+    def __init__(self, config=None, swagger_url=None, **session_kwargs):
```
hannes-ucsc: What's the advantage of having this parameter point at the `swagger.json` instead of just the base url (the on ending in `/v1`)?

test/integration/dss/test_dss_api.py
```diff
@@ -20,15 +21,41 @@
...
+    from unittest import mock
     TemporaryDirectory = tempfile.TemporaryDirectory
 
+
 class TestDssApi(unittest.TestCase):
     staging_bucket = "org-humancellatlas-dss-cli-test"
 
+    def test_set_host(self):
+        with TemporaryDirectory() as home:
+            with mock.patch.dict(os.environ, HOME=home):
```
hannes-ucsc: Since I wrote these tests, I noticed that `tweak` support `XDG_CONFIG_HOME`. Patching that instead of HOME would be more surgical.

hca/util/__init__.py
```diff
@@ -236,13 +237,12 @@ def store_refs(d):
     @property
     def swagger_spec(self):
         if not self._swagger_spec:
```
hannes-ucsc: I wonder if it would be beneficial to guard this section with a lock here so multiple threads racing through this at the same time don't all download the remote spec? It's a common pattern in my code to download the files in a bundle in parallel. All those threads start at the typically same time and go through the client initialization at the same time.

test/integration/dss/test_dss_api.py
```diff
@@ -20,15 +21,41 @@
...
+                dev = hca.dss.DSSClient(
+                    swagger_url="https://dss.dev.data.humancellatlas.org/v1/swagger.json")
+                self.assertEqual("dss.dev.data.humancellatlas.org", dev._swagger_spec['host'])
+
+    def test_set_host_multithreaded(self):
```
hannes-ucsc: Do these two tests cover all the cases I brought up in the ticket?

hca/util/__init__.py
calvinnhieu: In order to be consistent with the tweak config which uses `swagger_url`. Also, I think this makes more sense since:
- the true API host is defined in the swagger definition served at `swagger_url`
- `swagger_url` is used to generate the filename of the cached swagger definition. I think that exposing `swagger_url` makes more sense for developers.

test/integration/dss/test_dss_api.py
calvinnhieu: I would say so. These tests cover:
- specifying a "host" via `swagger_url` on client initialization
- ensuring that the host is set appropriately for a given `swagger_url` and does not revert to prod when multiple clients are initialized across multiple threads

Redefining the client host after initialization would require undoing and redoing the code generation of the API bindings for which we could just create a new client instead. The changes also no longer require the developer to use the HCAConfig workaround.

hca/util/__init__.py
calvinnhieu: That might save some time, but how much is debatable as the swagger definition is relatively small (<100kb) and the simultaneous fetch would only occur if the definition was not already cached + the time incurred synchronizing access to this property. Either way, I would say that it's out of the scope of this ticket and requires further investigation.

hca/util/__init__.py
hannes-ucsc: Makes sense.

hca/util/__init__.py
hannes-ucsc: > out of scope

Not really. This ticket is about thread safety, so locks fall into that category. I don't know if you have used locks before but I'm talking a about a three line change here. One line to import threading locks, one line to initialize the lock instance and one line for the `with lock` context manager invocation.

> That might save some time

It won't save time as the downloads happen concurrently. It saves server round-trips and lambda invocations. Considering that it is only three lines of code, I think it's worth it.

test/integration/dss/test_dss_api.py
hannes-ucsc: Ok. Just to be sure, I will run my test suite again, minus the workarounds.

hca/util/__init__.py
```diff
@@ -235,29 +238,29 @@ def store_refs(d):
...
-                except OSError as e:
-                    if not (e.errno == errno.EEXIST and os.path.isdir(self.config.user_config_dir)):
-                        raise
-                res = self.get_session().get(swagger_url)
-                res.raise_for_status()
-                assert "swagger" in res.json()
-                fs.atomic_write(os.path.basename(swagger_filename), self.config.user_config_dir, res.content)
-            with open(swagger_filename) as fh:
-                self.__class__._swagger_spec = self.load_swagger_json(fh)
+        with self.__class__._swagger_spec_lock:
```
hannes-ucsc: I'm not a fan of `self.__class__.attribute`. It's mildly paranoid. Class attributes are accessible directly on self.

test/integration/dss/test_dss_api.py
hannes-ucsc: Were you planning on addressing this?

PR #181, opened 2018-09-06
fix 0-byte file download (#132)

### LINKED ISSUE #132: Download fails with 0-byte files
# Steps to reproduce

1. Upload a 0-byte file.
2. Attempt to download it

calvinnhieu:
Range HTTP requests fail and return response code 416 if the requested range is not satisfiable - e.g. requesting a range starting from 0-bytes on a 0-byte file.

Fixes #132

### DIFF
--- a/hca/dss/__init__.py
+++ b/hca/dss/__init__.py
@@ -113,6 +113,9 @@ class DSSClient(SwaggerClient):
             retries_left = num_retries
             hasher = hashlib.sha256()
             with open(file_path, "wb") as fh:
+                if file_['size'] == 0:
+                    logger.info("%s", "File {}: CREATED (empty). Stored at {}.".format(filename, file_path))
+                    continue
                 while True:
                     try:
                         response = self.get_file._request(


calvinnhieu:
We could, but I went with this option in case there were inconsistencies between the manifest data and the actual files (i.e. is the manifest data a reliable source of truth?). And yes this writes out an empty (captured in the existing tests).

calvinnhieu:
Thought about it - you're right, going to skip the request if file size is 0

hca/dss/__init__.py
```diff
@@ -113,65 +113,66 @@ def download(self, bundle_uuid, replica, version="", dest_name="",
...
-                while True:
-                    try:
-                        response = self.get_file._request(
-                            dict(uuid=file_uuid, version=file_version, replica=replica),
-                            stream=True,
-                            headers={
-                                'Range': "bytes={}-".format(fh.tell())
-                            },
-                        )
+                if file_['size'] > 0:
```
mweiden: nit: this code is getting too deeply nested. I'd change the flow control here such that it's flatter.  Example below.

```python
def download(self, ...
    ...
    for file_ in files.values():
        ...
        if file_['size'] > 0:
            _download_file(file_uuid, file_version, replica)
        ....

def _download_file(file_uuid, file_version, replica)
...
```

ttung:
> We could, but I went with this option in case there are inconsistencies between the manifest data and the actual files (i.e. is the manifest data a reliable source of truth?). And yes this writes out an empty (captured in the existing tests).

I think the manifest data is a pretty reliable source of truth.

hca/dss/__init__.py
calvinnhieu: Updated w implementation that doesn't nest another layer, but agree the function can be refactored/broken down. Choosing not to refactor now in favor of focusing on other Pilot Phase tickets. Thoughts?

PR #182, opened 2018-09-07
The life time of a token can be adjusted for authentication sessions.

Bento007:
By adding the kwarg token_expiration to a SwaggerClient request you can specify how long before a generated security token expires.


Bento007:
I like that idea. It's easier to understand.

PR #183, opened 2018-09-07
update to use custom JWT email claim

Bento007:
updating claims inserted into JWT for for secure requests using google service credentials.

connects to https://github.com/HumanCellAtlas/data-store/pull/1562


Bento007:
I agree

PR #184, opened 2018-09-11
adding a backoff factor for api calls.

Bento007:


Bento007:
> We never got around to setting the backoff factor and it is good to set it, but we need to make sure the `Retry-After` header supersedes it

kozbo:
Background information from slack conversation:

Kevin [7:05 PM]
Can someone else give this a try using the preview data?
`hca dss get-file --replica aws --uuid f379d8e5-1741-4992-a46f-692f352d81e2 > MantonBM4_HiSeq_8_S32_L002_R2_001.fastq.gz`

Note that your config file should be as follows:
`{"log_level": "INFO", "client_id": "foo", "DSSClient": {"swagger_url": "https://dss.data.humancellatlas.org/v1/swagger.json"}}` (edited)

My attempts fail every time with `too many redirects` and if I try to get it from gcp I get `OSError: [Errno 22] Invalid argument` (edited)


akislyuk [5:38 AM]
The immediate reason why this fails is that Retry-After support in DCP CLI was broken in https://github.com/HumanCellAtlas/dcp-cli/pull/184. It should be reverted, and a new release rolled. (edited)


akislyuk [5:53 AM]
The underlying file checkout is failing too. You can see this by getting the execution id from the redirect target (after fixing the error in the CLI retry logic) and plugging it into e.g. `aws stepfunctions get-execution-history --execution-arn arn:aws:states:us-east-1:109067257620:execution:dss-s3-copy-sfn-prod:bcc756e7-9fb4-49c8-88c1-fd7f11a15284`

The proximate cause for the checkout failing is an etag mismatch:
    ```{
            "executionFailedEventDetails": {
                "cause": "{\"errorType\": \"AssertionError\", \"stackTrace\": [[\"/var/task/domovoi/app.py\", 203, \"__call__\", \"result = handler(event, context)\"], [\"/var/task/domovoilib/dss/stepfunctions/s3copyclient/implementation.py\", 198, \"join\", \"assert composite_etag == state[_Key.SOURCE_ETAG]\"]]}",
                "error": "AssertionError"
            },
            "timestamp": 1542202628.64,
            "type": "ExecutionFailed",
            "id": 369,
            "previousEventId": 368
        }```

I haven’t dug into the reason for the etag mismatch though.

PR #186, opened 2018-09-12
pin package

parthshahva:


requirements.txt
```diff
@@ -14,4 +14,4 @@ argcomplete >= 1.9.3, < 2
 commonmark >= 0.7.4, < 0.8
 docutils >= 0.14, < 1
 PyJWT >= 1.6.4
-retrying
+retrying==1.3.3
```
MDunitz: @kislyuk  had some concerns about using this package in the query service "If possible, we should stick to configuring the retry logic already embedded in our dependencies, instead of bringing in another one (I've never used retrying, but from a brief glimpse at its issue tracker, it appears unmaintained, with the last release in 2014 and a fork that is recommended to replace it)." I'm not sure if those concerns apply here

kislyuk:
In general, pinning library dependencies is not recommended because libraries or client-side packages in general (which dcp-cli is) are likely to be installed together with other packages, which may place their own version constraints on dependencies. If the different pins conflict, then some of them won't be satisfied. This is why it's important for libraries to have relaxed dependency version constraints (and ideally to test against multiple commonplace versions of dependencies), while applications can pin (freeze) all their dependency (and sub-dependency) versions. A good compromise between stability and flexibility is to depend on a major version range, for example: `foobar >= 1.2.3, < 2`.

Specifically regarding this package, please see #194

parthshahva:
Ok thanks @kislyuk. This makes sense. I also don't have any opposition to your statements against using the retrying module in the linked PR. I will remove the dependency as per the issue you created.

PR #187, opened 2018-09-17
US 169 Add command line tool for checking upload area or file status

MDunitz:
# Need
* Developers need to be able to check the status of a single file, or all files in an upload area

# Approach
* add command to check the status of an individual file (status) or upload area (report) via the command line
* status: pass in deployment stage, upload area and filename -> print the status of the file
* report: pass in deployment stage, upload area, optional name for generated file --> generate a file (named after the upload area if no filename arg included) listing the statuses of all files in the specified upload area

# Test
* unittests
* tested locally


hca/upload/api_client.py
```diff
@@ -32,6 +32,52 @@ def credentials(self, area_uuid):
...
+        url = "{api_url_base}/area/{uuid}/{filename}/checksum".format(api_url_base=self.api_url_base, uuid=area_uuid,
+                                                                      filename=filename)
+        response = requests.get(url)
+        if not response.ok:
+            raise RuntimeError(
+                "GET {url} returned {status}, {content}".format(
+                    url=url,
+                    status=response.status_code,
+                    content=response.content))
+        return response.json()
```
sampierson: The methods in this file have become very repetitive.  Time to dry them up.  It seems to me you could factor out a _get() and _post() method and delete half this file.

hca/upload/cli/list_file_status_command.py
```diff
@@ -0,0 +1,28 @@
...
+class ListFileStatusCommand(UploadCLICommand):
+    """
+    Print status of file in an upload area
+    """
+    @classmethod
+    def add_parser(cls, upload_subparsers):
+        list_file_statuses_parser = upload_subparsers.add_parser('status', help=cls.__doc__,
+                                                                 description=cls.__doc__)
+        list_file_statuses_parser.set_defaults(entry_point=ListFileStatusCommand)
+        list_file_statuses_parser.add_argument('env', help="Environment the upload area was created in")
```
sampierson: Could not perhaps the env and uuid be optional arguments, and default to the env configured in `.config/hca/config.json` and the current upload area, respectively?

I imagine that the usual usage of this file will be to check the status of files you just uploaded.  In that case having to specify env and area is overkill.

hca/upload/cli/generate_status_report_command.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    Generate file status report for upload area
+    """
+    @classmethod
+    def add_parser(cls, upload_subparsers):
+        gen_file_status_report_parser = upload_subparsers.add_parser('report', help=cls.__doc__,
+                                                                     description=cls.__doc__)
+        gen_file_status_report_parser.set_defaults(entry_point=GenerateStatusReportCommand)
+        gen_file_status_report_parser.add_argument('env', help="Environment the upload area was created in")
+        gen_file_status_report_parser.add_argument('uuid', help="Full UUID of an upload area.")
```
sampierson: Could not perhaps the env and uuid be optional arguments, and default to the env configured in .config/hca/config.json and the current upload area, respectively?

I imagine that the usual usage of this file will be to check the status of files you just uploaded. In that case having to specify env and area is overkill.

MDunitz: 👍

hca/upload/api_client.py
sampierson: CodeClimate can be a bit overzealous :)  Eliminating **every** *similar* piece of code would go way past the point of diminishing returns.  Just consider its feedback suggestions of where to look.

kislyuk:
FYI, this was merged with a codecov coverage decrease alert. Was there a problem with the coverage instrumentation? If not, it would be great not to reduce our code coverage.

Issue #179, opened 2018-08-31
Read The Docs TOC is duplicated


kozbo:
The formatting of the left hand TOC is inconsistent and has duplicated sections

see [docs](https://hca.readthedocs.io/en/latest/api.html)

![Screen Shot 2018-08-31 at 9.28.07 AM.png](https://images.zenhubusercontent.com/5a663ecd4b5806bc2bc3cc9c/7dbc8b1a-e1d8-4479-929c-b614481d8454)

Bento007:
#196 fixes this

PR #197, opened 2018-09-28
Fixing read the docs error with markup language in argsparser

Bento007:
*  The swagger doc that is used to generate the cli argparser was failing to render using sphinx-argparse because of markdown in the parameter descriptions. This was fixed by telling sphinx-argparser to expect markups.
* adding css for text wrapping long parameter descriptions in read the docs.

### DIFF
--- a/docs/_static/text_wrapping.css
+++ b/docs/_static/text_wrapping.css
@@ -1,0 +1,6 @@
+.wy-table-responsive table td {
+    white-space: normal !important;
+}
+.wy-table-responsive {
+    overflow: visible !important;
+}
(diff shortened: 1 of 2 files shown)


docs/_static/text_wrapping.css
```diff
@@ -0,0 +1,6 @@
+.wy-table-responsive table td {
+    white-space: normal !important;
```
calvinnhieu: Is the `!important` required?

Bento007: honestly I don't know. I pulled it from the sphinx-argparse docs https://sphinx-argparse.readthedocs.io/en/stable/misc.html#text-wrapping-in-argument-tables

kozbo:
Does this also resolve #179?

Bento007:
@kozbo no this is an unrelated issues cause by updates to the dss.

PR #204, opened 2018-11-12
[CVE-2018-18074] Upgrade requests to 2.20.0

mweiden:
CVE-2018-18074 More information
moderate severity
Vulnerable versions: <= 2.19.1
Patched version: 2.20.0

The Requests package through 2.19.1 before 2018-09-14 for Python sends an HTTP
Authorization header to an http URI upon receiving a same-hostname
https-to-http redirect, which makes it easier for remote attackers to discover
credentials by sniffing the network.


test/integration/upload/cli/test_upload.py
```diff
@@ -60,7 +60,7 @@ def test_upload_with_dcp_type_option(self):
 
     @responses.activate
     @patch('hca.upload.s3_agent.S3Agent.upload_file')   # Don't actually try to upload
-    def test_no_transfer_acceleration_option_sets_up_botocore_config_correctly(self, upload_file_stub):
+    def test_no_transfer_acceleration_option_sets_up_botocore_config_correctly(self, *args):
```
mckinsel: Is this because of a flake8 complaint? I think it would follow the style in the rest of the tests more if the `@patch` decorator got moved to the `with` block and the arguments got removed entirely, so like
```python
with patch('hca.upload.s3_agent.S3Agent.upload_file'), \
         patch('hca.upload.s3_agent.Config', new=Mock(wraps=botocore.config.Config)) as mock_config:
```

requirements-dev.txt
```diff
@@ -1,5 +1,5 @@
-flake8
-moto
+flake8==3.5.0
+moto==1.3.3
```
mckinsel: Do we have a reason to pin `moto` all the way down to patch version?

test/integration/upload/cli/test_upload.py
mweiden: No, actually this is because of a runtime error. Other suggestions?

requirements-dev.txt
mweiden: Can confirm that this doesn't work if I unpin the patch version 😢. I think we should figure this out _after_ getting the security fix out

mckinsel: Okay sounds good

PR #208, opened 2018-11-20
change retry lib and improve output from hca upload tool during file upload

parthshahva:


parthshahva:
@sampierson Good point on the commit message. Does the current one look a little better?

PR #225, opened 2019-01-22
modified Error API 

amarjandu:
Added checks for AWS-Request-ID headers that are returned in responses


hca/dss/__init__.py
```diff
@@ -305,7 +308,7 @@ def upload(self, src_dir, replica, staging_bucket, timeout_seconds=1200):
                     try:
                         self.head_file(uuid=file_uuid, replica="aws", version=version)
                         break
-                    except SwaggerAPIException as e:
+                    except APIException as e:
```
Bento007: change back to `SwaggerAPIException`

PR #146, opened 2018-07-31
help info for parameters appears in cli for upload and download

Bento007:
# Release Notes
- Using sphinx-argparser to generate documentation for the CLI into cli.rst
- refactored docstrings to restructed text format
- TOC now populates in sphinx documentation
- added common links to epilog in sphinx documentation
- using sphinx RSTParser to parse docstrings into argparser
- spread out documentation across api.rst, readme.rst, cli.rst, and index.rst

# Known Issue
- the sphinx documentation **incorrectly** marks all commands generated from a swagger doc as a `classmethod`.

PR #227, opened 2019-02-05
Fix DSS API documentation

Bento007:
Instantiate DSSClient to generate the documentation for swagger API binding


kislyuk:
(Also, in re: #146, whenever changing the doc build process, a good practice is to diff the output of the old build vs. the new.)

PR #243, opened 2019-02-22
Relax boto range

xbrianh:
boto3 >= 1.7, < 2


xbrianh:
Tests pass locally. Travis appears to be suffering auth errors.

PR #246, opened 2019-02-26
Add a function to warn the user if the CLI version is out of date.

DailyDreaming:


DailyDreaming:
Merging this then.  Especially as it contains bug fixes and a fairly minor version check.

kislyuk:
This causes the CLI to query PyPI every time that it starts. It will also fail if PyPI is down.

Can this please be reverted until a better solution is made?

PR #247, opened 2019-02-26
Slash support

amarjandu:


hca/dss/util/__init__.py
```diff
@@ -1,3 +1,45 @@
+import os
+
+
 def separator_to_camel_case(separated, separator):
     components = separated.split(separator)
     return "".join(x.title() for x in components)
+
+
+def directory_builder(src_dir: str):
```
DailyDreaming: Typing (`: str`) is python 3 only and this repo needs to be py2+3.

DailyDreaming:
I made a comment in review, but I also wanted to mention a failing test that you'll hit if you fix it.  Someone wrote some tests that check the exact size of the readme file and break if it ever changes.  Should be fixed by #246 when that's merged.

hca/dss/util/__init__.py
amarjandu: :(  alright, changing, thanks!

amarjandu:
Currently tests are breaking due to testing against Prod, support for slashes and subscriptions have only gotten into dev/staging. Testing against dev allows tests to pass.

hca/version.py
```diff
@@ -1 +1 @@
-__version__ = '0.0.0'
+__version__ = '4.8.0'
```
DailyDreaming: Looks like this accidentally got in.  Should change this back to `0.0.0`.

test/integration/dss/test_dss_api.py
```diff
@@ -56,6 +57,26 @@ def f(_):
                         with ThreadPoolExecutor(num_threads) as tpe:
                             self.assertTrue(all(x is None for x in tpe.map(f, range(num_threads))))
 
+    def test_python_nested_bundle_upload_download(self):
+        bundle_path = os.path.join(TEST_DIR, "res", "nestedBundle")
```
DailyDreaming: Nice.  Good accounting for slashes on both systems.

hca/version.py
amarjandu: ! how did that get there, reverting, thanks!

hca/dss/__init__.py
```diff
@@ -255,19 +261,22 @@ def upload(self, src_dir, replica, staging_bucket, timeout_seconds=1200):
...
+        file_uuids, uploaded_keys, abs_file_paths = upload_to_cloud(files_to_upload, staging_bucket=staging_bucket,
+                                                                    replica=replica, from_cloud=False)
         for file_handle in files_to_upload:
             file_handle.close()
-        filenames = list(map(os.path.basename, uploaded_keys))
+        filenames = [object_name_builder(p, src_dir) for p in abs_file_paths]
         filename_key_list = list(zip(filenames, file_uuids, uploaded_keys))
 
         for filename, file_uuid, key in filename_key_list:
+            filename = filename.replace('\\', '/')  # for windows paths
```
DailyDreaming: Does this work with `C:\\path\path\file.ex`?  I think it might need to strip the drive name for Windows files.

mikebaumann:
Hi @amarjandu, @hannes-ucsc had rightly stated that when downloading a bundle, it should be possible to download only the files under a given directory (or directories). I think the existing options to download `--metadata-files` and `--data-files` should work for this, given they take a shell-style pattern and perform file globbing. Does this work?
I am still working on the review, but haven't yet seen tests of this.
Thanks!

mikebaumann:
Other "interesting" (i.e. _potentially_ problematic) cases to consider for Windows are:
* `download` with the `--dest-name` parameter, and
* `upload` with the `<upload_path>` parameter

Where the parameter values are locations on a different Windows drive than the current drive.
For example, users current working directory is on `C:`, but the parameter value is something like
`D:\\mydata\myproject\mybundle`

Is the necessary retention of the drive letter handled properly for both upload and download?

hca/dss/__init__.py
amarjandu: the upload command would not work in this case, as it's only designated for directories, however it does handle the case in which the drive letter is included; within a support function the lettering for the directory is stripped as needed, see [hca/dss/util/__init__.py object_name_builder](hca/dss/util/__init__.py)

amarjandu:
@mikebaumann @DailyDreaming  I did some basic testing with the python bindings to see how cases with absolute Windows paths such as `D:\\User\amar\test`as both a upload source and destination directory would behave, so far if used as [raw strings](https://docs.python.org/3/reference/lexical_analysis.html#literals)

I believe that as long as the os can work with the directory; there should not be issues from different disk mount locations. But i do agree that we should look at some additional testing options for windows support.

hca/dss/__init__.py
```diff
@@ -92,16 +94,20 @@ def download(self, bundle_uuid, replica, version="", dest_name="",
...
 
-            if not os.path.isdir(dest_name):
-                os.makedirs(dest_name)
+            intermediate_path, filename = os.path.split(filename)
+            if intermediate_path:
+                walking_dir = os.path.join(walking_dir, intermediate_path)
+            if not os.path.isdir(walking_dir):
+                os.makedirs(walking_dir)
 
             logger.info("File %s: Retrieving...", filename)
```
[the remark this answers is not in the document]
amarjandu: The filename at ln.109 would have the intermediate directories included as in the bundle_manifest, however there is logging at ln.123 would show both the intermediate directories, as well as the destination directory for the download.

PR #252, opened 2019-02-28
Added clientside checksumming

maniarathi:
Checksumming on the client-side is "tagged" as a metadata to the S3 object.

Side note: I learned that crc32 is different than crc32c. The former is significantly faster than the latter but checksumming on a 1GB file upload took 1788 milliseconds to run which seems fine especially since the upload itself takes much longer.


sampierson:
Please could you test your solution from a local computer using a 1TB file and see how much time it adds to the transfer.  1GB files are not the problem, 5TB files are the problem.  Image files are going to be consistently over a terabyte.  

I'm wonder if there is a way to checksum the file as it is being uploaded by streaming it through BufferedReader or something. e.g. https://github.com/boto/boto3/issues/256#issuecomment-139609091

sampierson:
Aside: I believe crc32c was originally selected as one of the 4 DSS Checksums as that is what Google Cloud Storage automatically computes during upload.  The DSS may use that value somewhere.

maniarathi:
re: TB-sized files. These files would (I assume) follow the path of being copied over from an existing s3 area? Do you think they would likely be uploaded from a local path? The reason I ask is because in this PR, it doesn't checksum on S3 copies (next PR).

re: crc32c. How about sha1? It runs faster than crc32c anyways but will have to check. sha1 will still be pretty slow though for TB-sized files. What I might actually recommend is crc32 which isn't as secure but generally is good as a check to make sure the file wasn't corrupted during the actual upload. It's also blazingly fast (much much fast than any of the existing checksums). This does mean though that we'd have to add this server-side as well.

parthshahva:
On initial thought, using crc32 on client side checksumming and adding this to server side checksumming sounds reasonable. This allows us to confirm corruption-free uploads without affecting the 4 checksum tags that the dss is expecting. We can use s3 metadata instead of tags for this 5th checksum. We can then use this new crc32 metadata value in managing the syncing of s3/local directories to staging areas

maniarathi:
Ok, replaced with crc32 :) Will continue work on adding crc32 checksumming on serverside and then doing the crosscheck in separate PR.

hca/upload/checksum_handler.py
```diff
@@ -0,0 +1,67 @@
...
+import time
+import zlib
+
+import crcmod
+
+# Checksum(s) to compute for file; current options: crc32, crc32c, sha1, sha256
+CHECKSUM_NAMES = ['crc32']
+
+
+class ChecksumHandler:
```
Bento007: I'd recommend modifying and using https://github.com/HumanCellAtlas/dcplib/tree/master/dcplib/checksumming_io since it is shared across DCP components and will provide a uniform checksumming.

maniarathi: Done. https://github.com/HumanCellAtlas/dcplib/pull/6. Will modify this PR afterwards.

kozbo:
Is there an Issue that this PR relates to? If so then please connect it.

maniarathi:
Ready for re-review.

This PR is related to an open upload-service issue: https://app.zenhub.com/workspaces/dcp-backlogs-5ac7bcf9465cb172b77760d9/issues/humancellatlas/upload-service/37

hca/upload/client_side_checksum_handler.py
```diff
@@ -0,0 +1,53 @@
...
+class ClientSideChecksumHandler:
+    """ The ClientSideChecksumHandler takes in a file as a parameter and handles any behavior related to
+    check-summing the file
+    on the client-side, returning a tag that can be used as metadata when the file is uploaded to S3."""
+
+    def __init__(self, filename):
+        self._checksums = {}
+        self._filename = filename
+
+    def getChecksumMetadataTag(self):
```
parthshahva: Any reason for camel case vs standard python underscore naming?

parthshahva: https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

hca/upload/s3_agent.py
```diff
@@ -83,7 +83,8 @@ def copy_s3_file(self, s3_path, target_bucket, target_key, content_type, report_
             'ExtraArgs': {
                 'ContentType': content_type,
                 'MetadataDirective': 'REPLACE',
-                'ACL': 'bucket-owner-full-control'
+                'ACL': 'bucket-owner-full-control',
+                'Metadata': checksums
```
parthshahva: Does this stick in s3? I remember I had some difficulties earlier with s3 metadata

hca/upload/client_side_checksum_handler.py
maniarathi: Nope... just habit.

hca/upload/s3_agent.py
maniarathi: Yep! Tried and tested!

parthshahva:
Also another question: Did we cut a new version of dcplib that this PR depends on or does a new version of dcplib still need to be cut?

maniarathi:
New version has already been cut and the versioning in requirements.txt has been accordingly updated.

DailyDreaming:
@maniarathi Please be more careful in the future with force-pushes in the future as this PR reverted a previous PR: https://github.com/HumanCellAtlas/dcp-cli/pull/281

maniarathi:
Oh no! Sorry about that @DailyDreaming

DailyDreaming:
No worries!  Just be mindful!

Issue #219, opened 2018-12-18
CLI Help Menu Formatting


amarjandu:
As a user of the CLI I would like to quickly look over the help menus for commands to get more information about the command that will be ran. 

# Problem

Currently the CLI help menu is missing information about what input the commands can utilize, this includes information about required/optional parameters, expected formatting for certain parameters.

This causes issues with having to branch off and find scattered information through READMEs or on swagger API pages.

# Reproduction 

```
 [venv-data-store]amar:~/dev/testing/hca-cli! hca dss get-bundles-checkout -h 
usage: hca dss get-bundles-checkout [-h] --replica {aws,gcp} --checkout-job-id
                                    CHECKOUT_JOB_ID

Use this route with the `checkout_job_id` identifier returned by `POST
/bundles/{uuid}/checkout`.

optional arguments:
  -h, --help            show this help message and exit
  --replica {aws,gcp}   Replica to fetch from.
  --checkout-job-id CHECKOUT_JOB_ID
                        A RFC4122-compliant ID for the checkout job request.

[venv-data-store]amar:~/dev/testing/hca-cli! hca dss get-bundles-checkout
usage: hca dss get-bundles-checkout [-h] --replica {aws,gcp} --checkout-job-id
                                    CHECKOUT_JOB_ID
hca dss get-bundles-checkout: error: argument --replica is required

```
--replica is identified as an optional, then later as a requirement 

```
[venv-data-store]amar:~/dev/testing/hca-cli$ hca dss put-bundle -h
usage: hca dss put-bundle [-h] --version VERSION --replica {aws,gcp} --uuid
                          UUID --files FILES [FILES ...] --creator-uid
                          CREATOR_UID

Create a new version of a bundle with a given UUID. The list of file
UUID+versions to be included must be provided.

optional arguments:
  -h, --help            show this help message and exit
  --version VERSION     Timestamp of bundle creation in DSS_VERSION format.
  --replica {aws,gcp}   Replica to write to.
  --uuid UUID           A RFC4122-compliant ID for the bundle.
  --files FILES [FILES ...]
  --creator-uid CREATOR_UID
                        User ID who is creating this bundle.

```

`--files` does not have a description for how the input should be specified to the user.

DailyDreaming:
I was just thinking of proposing making this a sphinx project, when Lo and Behold, I Gaze Upon It and It Is There.

https://github.com/HumanCellAtlas/dcp-cli/pull/146

[...]

PR #254, opened 2019-03-02
Properly place and print required vs. non-required args in the help menu.  Addresses #219.

DailyDreaming:
Addresses #219.

This PR separates the args into `Positional Arguments`, `Required Arguments`, and `Optional Arguments` based on whether the swagger (or internal CLI function) lists them as required or not.

So this only groups arguments differently.

For example, currently it prints:
```
(v3nv) quokka@qcore:~/git/dcp-cli$ hca dss get-file -h
usage: hca dss get-file [-h] --uuid UUID --replica {aws,gcp}
                        [--version VERSION] [--token TOKEN]

Given a file UUID, return the latest version of that file. If the version is
provided, that version of the file is returned instead. Headers will contain
the data store metadata for the file. This endpoint returns a HTTP redirect to
another HTTP endpoint with the file contents.

optional arguments:
  -h, --help           show this help message and exit
  --version VERSION    Timestamp of file creation in DSS_VERSION format. If
                       this is not provided, the latest version is returned.
  --token TOKEN        Token to manage retries. End users constructing queries
                       should not set this parameter.
  --uuid UUID          A RFC4122-compliant ID for the file.
  --replica {aws,gcp}  Replica to fetch from.
```
This would change to:
```
(v3nv) quokka@qcore:~/git/dcp-cli$ hca dss get-file -h
usage: hca dss get-file [-h] --uuid UUID --replica {aws,gcp}
                        [--version VERSION] [--token TOKEN]

Given a file UUID, return the latest version of that file. If the version is
provided, that version of the file is returned instead. Headers will contain
the data store metadata for the file. This endpoint returns a HTTP redirect to
another HTTP endpoint with the file contents.

Optional Arguments:
  -h, --help           show this help message and exit
  --version VERSION    Timestamp of file creation in DSS_VERSION format. If
                       this is not provided, the latest version is returned.
  --token TOKEN        Token to manage retries. End users constructing queries
                       should not set this parameter.

Required Arguments:
  --uuid UUID          A RFC4122-compliant ID for the file.
  --replica {aws,gcp}  Replica to fetch from.
```


hca/util/__init__.py
```diff
@@ -83,9 +83,15 @@ def special_cli_command(self, required_argument, optional_argument=""):
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import os, types, collections, typing, json, errno, base64, argparse
+import os
```
Bento007: Why?

DailyDreaming: PEP8: https://www.python.org/dev/peps/pep-0008/#imports

DailyDreaming:
@Bento007 Should be addressed.  Example output now copy-pasted at the top.

hca/util/__init__.py
```diff
@@ -520,13 +526,15 @@ def _get_param_argparse_type(self, anno):
     def build_argparse_subparsers(self, subparsers):
         for method_name, method_data in self.methods.items():
             subcommand_name = method_name.replace("_", "-")
-            subparser = subparsers.add_parser(subcommand_name, help=method_data.get("summary"),
-                                              description=method_data.get("description"))
+            optional_args = subparsers.add_parser(subcommand_name, help=method_data.get("summary"),
+                                                  description=method_data.get("description"))
+            required_args = optional_args.add_argument_group('Required Arguments')
```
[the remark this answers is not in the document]
DailyDreaming: Ah, good point.  I can always use better naming, thanks!

.gitignore
```diff
@@ -42,3 +42,7 @@ npm-debug.log
 
 # Credential files
 /gcs-credentials.json
+
+# virtualenv
+venv/
```
Bento007: this should be in your global .gitignore not in this repo.

DailyDreaming: How do you distinguish the two?

PR #257, opened 2019-03-05
301 retries are logged as info

### LINKED ISSUE #256: WARNING messages from the CLI
Talked to @hannes-ucsc and @NoopDog about this a bit already. Here are the WARNING messages I was seeing. I think the messages were just a little bit overwhelming at first and that's what threw me off. 

hcaclierrormessages

Bento007:
Fixes #256

### DIFF
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -115,11 +115,12 @@ class RetryPolicy(retry.Retry):
 
     def increment(self, *args, **kwargs):
         _retry = super(RetryPolicy, self).increment(*args, **kwargs)
-        last_resp = retry.history[-1]
+        last_resp = _retry.history[-1]
         if last_resp.status in {301}:
-            logger.info("Retrying: {}".format(last_resp))
+            log_lvl = logger.info
         else:
-            logger.warning("Retrying: {}".format(last_resp))
+            log_lvl = logger.warning
+        log_lvl("Retrying: {}".format(last_resp))
         return _retry
 
 
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -114,9 +114,13 @@ class RetryPolicy(retry.Retry):
         self.RETRY_AFTER_STATUS_CODES = frozenset(retry_after_status_codes | retry.Retry.RETRY_AFTER_STATUS_CODES)
 
     def increment(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


hca/util/__init__.py
```diff
@@ -114,9 +114,13 @@ def __init__(self, retry_after_status_codes={301}, *args, **kwargs):
         self.RETRY_AFTER_STATUS_CODES = frozenset(retry_after_status_codes | retry.Retry.RETRY_AFTER_STATUS_CODES)
 
     def increment(self, *args, **kwargs):
-        retry = super(RetryPolicy, self).increment(*args, **kwargs)
-        logger.warning("Retrying: {}".format(retry.history[-1]))
-        return retry
+        _retry = super(RetryPolicy, self).increment(*args, **kwargs)
+        last_resp = retry.history[-1]
```
hannes-ucsc: Can retry.history be empty?

hca/util/__init__.py
```diff
@@ -114,9 +114,13 @@ def __init__(self, retry_after_status_codes={301}, *args, **kwargs):
...
+        if last_resp.status in {301}:
+            logger.info("Retrying: {}".format(last_resp))
```
hannes-ucsc: Might want to reduce the duplication between those two lines.

hca/util/__init__.py
Bento007: not at this point in the code because `super(RetryPolicy, self).increment(*args, **kwargs)` appends the last request to history.

hca/util/__init__.py
```diff
@@ -114,9 +114,14 @@ def __init__(self, retry_after_status_codes={301}, *args, **kwargs):
         self.RETRY_AFTER_STATUS_CODES = frozenset(retry_after_status_codes | retry.Retry.RETRY_AFTER_STATUS_CODES)
 
     def increment(self, *args, **kwargs):
-        retry = super(RetryPolicy, self).increment(*args, **kwargs)
-        logger.warning("Retrying: {}".format(retry.history[-1]))
-        return retry
+        _retry = super(RetryPolicy, self).increment(*args, **kwargs)
+        last_resp = _retry.history[-1]
+        if last_resp.status in {301}:
```
hannes-ucsc:
```suggestion
        if last_resp.status == 301:
```

hannes-ucsc: And I would use a ternary expression.

Issue #222, opened 2018-12-21
`iterate` methods don't handle errors very well


mckinsel:
When using the python API, I often have to use a pattern that looks like this

```python
while True:
    try:
        # <something with the DSS>
        break
    except:
        continue
```

This is because requests to the DSS often fail for any number of reasons and succeed later on retry. But I don't think you can use such a pattern with `iterate`. So if you get, say, a random 504, you have to start all over again.

kislyuk:
A representative error showing this issue:
```
Traceback (most recent call last):
  File "./fetch.py", line 65, in <module>
    print(len(list(dss.post_search.iterate(es_query={}, replica="aws"))))
  File "/usr/local/lib/python3.6/dist-packages/hca/util/__init__.py", line 183, in iterate
    page = self._request(kwargs, url=page.links["next"]["url"] if page else None)
  File "/usr/local/lib/python3.6/dist-packages/hca/util/__init__.py", line 149, in _request
    raise SwaggerAPIException(response=res)
hca.util.exceptions.SwaggerAPIException: unhandled_exception: sequence item 0: expected str instance, NoneType found (HTTP 500). Details:
Traceback (most recent call last):
  File "/var/task/chalicelib/dss/error.py", line 55, in wrapper
    return func(*args, **kwargs)
  File "/var/task/chalicelib/dss/api/search.py", line 80, in post
    next_url = _build_next_url(page, per_page, replica_enum, output_format)
  File "/var/task/chalicelib/dss/api/search.py", line 150, in _build_next_url
    search_after = ','.join(page['hits']['hits'][-1]['sort'])
TypeError: sequence item 0: expected str instance, NoneType found
```
Here `dss` is an instance of `hca.dss.DSSClient()`.

I believe this is caused by the iterator internals not installing standard retry logic in pagination requests.

kislyuk:
Digging into this a little bit deeper, I think this is not specific to the iterator, but is instead because we don't retry at all on 500 errors. Do we have a rationale why? We should probably go ahead and add all 500 series codes to hca.util.RetryPolicy.

DailyDreaming:
See #267 .

PR #265, opened 2019-03-07
 Add 500 series error codes to the retry policy.

DailyDreaming:
Addresses #222 and based on the comments there.

### DIFF
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -115,9 +115,19 @@ from .fs_helper import FSHelper as fs
 
 
 class RetryPolicy(retry.Retry):
-    def __init__(self, retry_after_status_codes={301}, *args, **kwargs):
+    def __init__(self,
+                 retry_after_status_codes={301},
+                 read=10,
+                 status=10,
+                 backoff_factor=0.1,
+                 status_forcelist=frozenset(set(range(500, 512))),
+                 *args, **kwargs):
         super(RetryPolicy, self).__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


hca/util/__init__.py
```diff
@@ -115,9 +115,19 @@ def special_cli_command(self, required_argument, optional_argument=""):
 
 
 class RetryPolicy(retry.Retry):
-    def __init__(self, retry_after_status_codes={301}, *args, **kwargs):
+    def __init__(self,
+                 retry_after_status_codes={301},
+                 read=10,
+                 status=10,
+                 backoff_factor=0.1,
+                 status_forcelist=frozenset(set(range(500, 512))),
```
xbrianh: let's preserve the original set for now: `{500, 502, 503, 504}`

DailyDreaming: Sounds good.  Will close this until I have something to test with.

hca/util/__init__.py
```diff
@@ -115,9 +115,19 @@ def special_cli_command(self, required_argument, optional_argument=""):
 
 
 class RetryPolicy(retry.Retry):
-    def __init__(self, retry_after_status_codes={301}, *args, **kwargs):
+    def __init__(self,
+                 retry_after_status_codes={301},
+                 read=10,
+                 status=10,
+                 backoff_factor=0.1,
```
xbrianh: Since we're not sure what we're fixing, perhaps we should adopt a conservative philosophy. I suggest leaving `RetryPolicy` unchanged except for `status_forcelist`.

Issue #261, opened 2019-03-06
Enhance DCP CLI DSS upload and download to support "/" in bundle file names


mikebaumann:
See: Proposal to add support for “/” virtual paths inside bundles https://github.com/HumanCellAtlas/data-store/issues/1885

PR #266, opened 2019-03-07
Slash-Support : supporting '/' within upload/download for files in bundles

amarjandu:
Supporting `/` within upload/download for files in bundles

Address the need to support nested folders/files within bundle
structures.
Imaging data bundles will have specifc file structures that neeed to
be preserved the same when being uploaded or downloaded to the DSS.
This also addresses a workaround for the matrix team where `!` was
being used due to not having slash-support

Resolves: #261
See also: https://github.com/HumanCellAtlas/data-store/issues/1885


hca/cli.py
```diff
@@ -112,7 +112,7 @@ def main(args=None):
     else:
         parser = get_parser()
 
-    if len(sys.argv) < 2:
+    if args and len(args) < 2:
```
DailyDreaming: Please change:
`if args and len(args) < 2:`

to:

```
if not args:
    args = sys.argv

if len(args) < 2:
```

hca/cli.py
```diff
@@ -111,8 +111,9 @@ def main(args=None):
         parser = get_parser(help_menu=True)
     else:
         parser = get_parser()
-
-    if len(sys.argv) < 2:
+        if not args:
```
DailyDreaming: Indenting.

amarjandu: changed, thx

hca/cli.py
```diff
@@ -107,12 +107,14 @@ def help(args):
 
 
 def main(args=None):
+    if not args:
+        args = sys.argv
     if '--help' in sys.argv or '-h' in sys.argv:
```
DailyDreaming: Could you change these refs to `sys.argv` to `args`?

amarjandu: addressed

test/integration/dss/test_dss_cli.py
```diff
@@ -31,24 +33,32 @@ def test_get_files_cli(self):
...
 
-        self.assertIsInstance(res, bytes)
+        with open(os.path.join(dest_dir, filename), 'rb') as download_data:
+            download_content = download_data.read()
         with open(file_path, "rb") as bytes_fh:
             file_content = bytes_fh.read()
-            self.assertEqual(file_content, res)
+            self.assertEqual(file_content, download_content)
+
+        shutil.rmtree(dest_dir)
```
mikebaumann: As it stands now, this temp directory will not be cleaned-up if the test execution does not reach this point.
I propose putting the test body in a `try` block with the `finally` clause containing the removal of the directory.

amarjandu: good catch, ill implement that now

hca/dss/util/__init__.py
```diff
@@ -1,3 +1,39 @@
+import os
+from scandir import scandir
+
+
 def separator_to_camel_case(separated, separator):
     components = separated.split(separator)
     return "".join(x.title() for x in components)
+
+
+def directory_builder(src_dir):
```
xbrianh: Consider renaming this to something like `iter_paths`

mikebaumann:
@amarjandu Assuming you are going to squash and merge this, please:
* Add a more descriptive title to the PR
* A more complete description of the change in the PR
* A link to the associated issue
Thanks!

hca/dss/util/__init__.py
```diff
@@ -1,3 +1,39 @@
...
+
+
+def object_name_builder(file_name, src_dir):
+    """
+    Function creates a name to be uploaded into the manifest
+    :param src_dir: string for src directory, used for removing path information
+    :param file_name: filename string to be cleaned
+    :return: returns a name to be used for the cloud object
+    """
+    file_path = os.path.normpath(os.path.join(file_name))
```
xbrianh: Is `file_name` expected to be a list, or list-like?

hca/dss/util/__init__.py
```diff
@@ -1,3 +1,39 @@
...
+    """
+    for x in scandir(os.path.join(src_dir)):
+        if x.is_dir(follow_symlinks=False):
+            for x in directory_builder(x.path):
+                yield x
+        else:
+            yield x
+
+
+def object_name_builder(file_name, src_dir):
```
xbrianh: Can you explain the difference between this implementation and:
```
def object_name_builder(file_name, src_dir):
    return os.path.normpath(os.path.join(file_name)).replace(src_dir, "")
```

hca/dss/util/__init__.py
amarjandu: file_name would be a single str object

hca/dss/util/__init__.py
amarjandu: i like your implementation more, ill refactor.

hca/dss/__init__.py
```diff
@@ -92,16 +93,20 @@ def download(self, bundle_uuid, replica, version="", dest_name="",
...
 
-            if not os.path.isdir(dest_name):
-                os.makedirs(dest_name)
+            intermediate_path, filename = os.path.split(filename)
+            if intermediate_path:
+                walking_dir = os.path.join(walking_dir, intermediate_path)
+            if not os.path.isdir(walking_dir):
+                os.makedirs(walking_dir)
 
             logger.info("File %s: Retrieving...", filename)
```
mikebaumann: I think this should log the full filename, including the path, as it is in the bundle.
I think this line:
```
 intermediate_path, filename = os.path.split(filename)
```
should be changed to something like:
```
 intermediate_path, filebasename = os.path.split(filename)
```
Rather than having `filename` representing the full path at one point in the code, then being changed to just the basename at some point, I think it would be better to introduce a new name.
If this is done, the the original `filename` (with the path, if present) would be the correct value to log.

amarjandu: implimented

amarjandu: done

hca/cli.py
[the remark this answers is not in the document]
amarjandu: done

hca/cli.py
```diff
@@ -107,12 +107,14 @@ def help(args):
...
         parser = get_parser(help_menu=True)
     else:
         parser = get_parser()
 
-    if len(sys.argv) < 2:
+    if len(args) < 2:
```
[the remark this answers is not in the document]
amarjandu: i added this change but needed to also perform
```   
 if not args:
       args = sys.argv[1:]
```
this resolves a parsing issue with the script name

hca/cli.py
```diff
@@ -107,12 +107,14 @@ def help(args):
 
 
 def main(args=None):
-    if '--help' in sys.argv or '-h' in sys.argv:
+    if not args:
+        args = sys.argv[1:]
```
amarjandu: @DailyDreaming i had to do this to get around parser issue when trying to use the cli after performing a `make install` everything works within testing + manual cli usage. Can you think of a reason it would break a parser later on?

DailyDreaming: No, I think that was the correct call.

PR #267, opened 2019-03-07
[Easy] Temporary fix to allow POST requests to be retried.

DailyDreaming:
Addresses #222 .

Iterate wasn't retrying on this:
```
from hca import HCAConfig
from hca.dss import DSSClient

hca_config = HCAConfig()
hca_config['DSSClient'].swagger_url = f'https://dss.dev.data.humancellatlas.org/v1/swagger.json'
dss = DSSClient(config=hca_config)
print(len(list(dss.post_search.iterate(es_query={}, replica="aws"))))
```
And so would break on a 500, even though we should retry on 500 errors.

That's because `urllib3.util.retry` doesn't retry on POST requests.  This fixes that temporarily.

The longer term solution is here: https://github.com/HumanCellAtlas/data-store/pull/1961  See comments there for follow-up steps.

We'll have to wait a month+ for that to percolate up to `prod` though, so this is a stop-gap in the meantime.

### DIFF
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -115,9 +115,15 @@ from .fs_helper import FSHelper as fs
 
 
 class RetryPolicy(retry.Retry):
-    def __init__(self, retry_after_status_codes={301}, *args, **kwargs):
+    DEFAULT_METHOD_WHITELIST = frozenset(['HEAD', 'GET', 'PUT', 'DELETE', 'OPTIONS', 'TRACE', 'POST'])
+
+    def __init__(self,
+                 retry_after_status_codes={301, 500},
+                 method_whitelist=DEFAULT_METHOD_WHITELIST,
+                 *args, **kwargs):
         super(RetryPolicy, self).__init__(*args, **kwargs)
         self.RETRY_AFTER_STATUS_CODES = frozenset(retry_after_status_codes | retry.Retry.RETRY_AFTER_STATUS_CODES)
+        self.method_whitelist = method_whitelist  # methods not on this list will not be retried
 
     def increment(self, *args, **kwargs):
         _retry = super(RetryPolicy, self).increment(*args, **kwargs)
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -118,7 +118,7 @@ class RetryPolicy(retry.Retry):
     DEFAULT_METHOD_WHITELIST = frozenset(['HEAD', 'GET', 'PUT', 'DELETE', 'OPTIONS', 'TRACE', 'POST'])
 
     def __init__(self,
-                 retry_after_status_codes={301, 500},
+                 retry_after_status_codes={301},
                  method_whitelist=DEFAULT_METHOD_WHITELIST,
                  *args, **kwargs):
         super(RetryPolicy, self).__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


DailyDreaming:
@mckinsel Does this solve your problem?

DailyDreaming:
@kislyuk Thank you.  I'll look into implementing just that.

hca/util/__init__.py
```diff
@@ -115,9 +115,15 @@ def special_cli_command(self, required_argument, optional_argument=""):
...
-    def __init__(self, retry_after_status_codes={301}, *args, **kwargs):
+    DEFAULT_METHOD_WHITELIST = frozenset(['HEAD', 'GET', 'PUT', 'DELETE', 'OPTIONS', 'TRACE', 'POST'])
+
+    def __init__(self,
+                 retry_after_status_codes={301},
+                 method_whitelist=DEFAULT_METHOD_WHITELIST,
+                 *args, **kwargs):
         super(RetryPolicy, self).__init__(*args, **kwargs)
         self.RETRY_AFTER_STATUS_CODES = frozenset(retry_after_status_codes | retry.Retry.RETRY_AFTER_STATUS_CODES)
+        self.method_whitelist = method_whitelist  # methods not on this list will not be retried
```
mckinsel: Should this instead be in the `retry.Retry.__init__` call?

DailyDreaming: Removed the whitelist usage.

DailyDreaming:
@kislyuk This is now a fix that implements a separate, basic (10) retry limit on POST search only.

I've tested it and it seems to resolve the problem.  It's intended to be temporary until the swagger doc changes in `data-store` percolate up to `prod` which should be at least a month from now I'm guessing.

xbrianh:
Can we indicate the temporary nature of this PR in the title?

DailyDreaming:
@xbrianh Done.

hca/util/__init__.py
```diff
@@ -131,11 +131,41 @@ def increment(self, *args, **kwargs):
 
 
 class _ClientMethodFactory(object):
-
     def __init__(self, client, parameters, path_parameters, http_method, method_name, method_data, body_props):
         self.__dict__.update(locals())
         self._context_manager_response = None
 
+    def request_with_retries_on_post_search(self, session, url, query, json_input, stream, headers):
```
xbrianh: Is this method used in `_PaginatingClientMethodFactory`?

DailyDreaming: It is.  `_PaginatingClientMethodFactory` calls `_request` which calls `request_with_retries_on_post_search`.

kislyuk:
I'm OK with this as long as you add a TODO comment in the code and commit description to revert this once the other PR is merged and released.

DailyDreaming:
@kislyuk Sounds good, will add it.

DailyDreaming:
@kislyuk I've added a TODO with a link to the swagger PR.

kislyuk:
What's the ETA for this to be reverted?

DailyDreaming:
@kislyuk I'll bring it up at stand up on Monday.  Should be next week some time.

DailyDreaming:
@kislyuk With the caveat that the deploy from `staging` to `prod` isn't delayed on Tuesday.

PR #274, opened 2019-03-14
Add version directory for manifest download (fixes #251)

### LINKED ISSUE #251: Atomically and reliably create a `.version` for a download file store
The version file should contain the layout version and its parameters. This makes it possible to update the layout in the future. The file needs to be created if it doesn't exist and that's where the race conditions come in. These race conditions should be handled. Note that it's not sufficient to just create the file atomically because that would mean that last writer wins.

jessebrennan:
This directory serves as a way to compartmentalize the downloaded files and enforces that different potential download schemes do not conflict.

### DIFF
--- a/hca/dss/__init__.py
+++ b/hca/dss/__init__.py
@@ -219,9 +219,10 @@ class DSSClient(SwaggerClient):
         :param checksum: a string checksum
         :return: relative Path object
         """
-        assert(sum(cls.DIRECTORY_NAME_LENGTHS) <= len(checksum))
+        file_prefix = '_'.join(['files'] + list(map(str, cls.DIRECTORY_NAME_LENGTHS)))
+        path_pieces = ['.hca', 'v2', file_prefix]
         checksum_index = 0
-        path_pieces = []
+        assert(sum(cls.DIRECTORY_NAME_LENGTHS) <= len(checksum))
         for prefix_length in cls.DIRECTORY_NAME_LENGTHS:
             path_pieces.append(checksum[checksum_index:(checksum_index + prefix_length)])
             checksum_index += prefix_length
--- a/test/unit/test_manifest_download.py
+++ b/test/unit/test_manifest_download.py
@@ -56,19 +56,20 @@ class TestManifestDownload(unittest.TestCase):
         files_present = {os.path.join(dir_path, f)
                          for dir_path, _, files in scandir.walk('.')
                          for f in files}
+        version_dir = os.path.join('.', '.hca', 'v2', 'files_2_4')
         files_expected = {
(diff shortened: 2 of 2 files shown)


hca/dss/__init__.py
```diff
@@ -269,10 +273,11 @@ def download_manifest_v2(self, manifest, replica, num_retries=10, min_delay_seco
         """
         rows = self._parse_manifest(manifest)
         errors = 0
+        download_version = 'v2'
```
hannes-ucsc:
```suggestion
```

hca/dss/__init__.py
```diff
@@ -269,10 +273,11 @@ def download_manifest_v2(self, manifest, replica, num_retries=10, min_delay_seco
...
         for row in rows:
             file_uuid = row['file_uuid']
             sha = row['file_sha256']
-            path = self._file_path(sha)
+            path = self._file_path(sha, version=download_version)
```
hannes-ucsc:
```suggestion
            path = self._file_path(sha)
```

hca/dss/__init__.py
```diff
@@ -30,6 +30,9 @@
 from .upload_to_cloud import upload_to_cloud
 
 
+DOWNLOAD_DIR_ROOT = '.hca'
```
hannes-ucsc:
```suggestion

```

hca/dss/__init__.py
```diff
@@ -213,15 +216,16 @@ def _download_file(self,
                 logger.info("%s", "File {}: GET SUCCEEDED. Stored at {}.".format(file_uuid, dest_path))
 
     @classmethod
-    def _file_path(cls, checksum):
+    def _file_path(cls, checksum, version='v2'):
```
hannes-ucsc:
```suggestion
    def _file_path(cls, checksum):
```

hca/dss/__init__.py
```diff
@@ -213,15 +216,16 @@ def _download_file(self,
...
         """
         returns a file's relative local path based on the nesting parameters and the files hash
         :param checksum: a string checksum
         :return: relative Path object
         """
-        assert(sum(cls.DIRECTORY_NAME_LENGTHS) <= len(checksum))
+        file_prefix = '_'.join(['files'] + list(map(str, cls.DIRECTORY_NAME_LENGTHS)))
+        path_pieces = [DOWNLOAD_DIR_ROOT, version, file_prefix]
```
hannes-ucsc:
```suggestion
        path_pieces = ['.hca', 'v2', file_prefix]
```

Keep all the bits in one place.

jessebrennan: Alright I'll make these changes.

Issue #260, opened 2019-03-05
Add Windows platform for dcp-cli CI


mikebaumann:
@kozbo commented on [Wed Feb 27 2019](https://github.com/HumanCellAtlas/data-store/issues/1922)

## User story
As a user of the HCS DSS CLI, I want the CLI to work correctly/reliably when run on Windows platforms

This is being done now in support of the concurrent work to 1) enable paths in bundle file names, and 2) improve dcp-cli download performance/efficiency, including through the use of hard links, etc.

## Definition of done
DCP CLI unit tests are automatically and successfully run on Windows, in addition to the existing Linux test platform, as part of continuous integration

---

@mikebaumann commented on [Thu Feb 28 2019](https://github.com/HumanCellAtlas/data-store/issues/1922#issuecomment-468370754)

The current plan consists of the following:

- [x] Move the current dcp-cli CI setup/execution from Travis to GitLab. The existing tests and CI test execution will remain the same, including testing of the four currently tested Python versions (2.7, 3.4, 3.5, 3.6), etc.
- [x] Enhance the pipeline above with additional job(s) to run the dcp-cli unit tests on Windows. This will at least be done for the latest supported version of Python (3.6), not sure if we want/need to test all versions of Python on both Linux and Windows. TBD, open to comment.
- [x] Setup Terraform script for the ec2 instance that holds the gitlab-windows-runner, this should live in the allspark repo.

xbrianh:
Some notes:
- Eventually we may be able to add a Windows node into the EKS cluster running GitLab CI/CD jobs:  https://github.com/aws/containers-roadmap/issues/69
- In the meantime, we should look into parameterizing an EC2 instance running windows, along with runner installation scripting, in the allspark repo: https://github.com/HumanCellAtlas/allspark

amarjandu:
## Notes 3/25/19 : 8:34 AM

### Complete
added sections to make file for windows build

### In Progress
secrets fetching mechanism

### TODO
run tests via make

### Awaiting 
terraform resources; add runner to cli

amarjandu:
## Notes 3/26/19 : 10:07 AM
docker within windows defaults to looking at the hyper-v subnet for AWS Metadata Server, have to adjust this for each docker instance created.

### Complete
ec2 + gitlab runner + docker working

### In Progress
Updating gitlab to get allspark -> ec2 windows runner communication,
see https://gitlab.com/gitlab-org/gitlab-runner/issues/3755

[...]

PR #278, opened 2019-03-16
GitLab CI configuration for Linux

mikebaumann:
GitLab CI configuration for Linux that tests
multiple versions of Python.

The code coverage results are not uploaded
to https://codecov.io/ as that is currently being done by Travis CI.

connects to https://github.com/HumanCellAtlas/dcp-cli/issues/260

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -34,9 +34,6 @@ stages:
   except:
     - tags
 
-after_script:
-    - bash <(curl -s https://codecov.io/bash)
-
 test-2.7:
   extends: .linux-test-base
   variables:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,70 @@
+variables:
+  AWS_DEFAULT_REGION: "us-east-1"
+  DSS_SECRETS_STORE: "dcp/dss"
+  DSS_TEST_STAGE: "dev"
+  BOTO_CONFIG: "/dev/null"
+
+stages:
+    - test
+
+.linux-test-base:
+  stage: test
+  image: ubuntu:18.04
+  before_script:
+    - IFS='.'; semver=($PYTHON_VERSION); PYTHON_MAJOR=${semver[0]}; unset IFS
+    - export DEBIAN_FRONTEND=noninteractive
+    - apt-get update --quiet
+    - apt-get install --assume-yes --no-install-recommends software-properties-common
+    - add-apt-repository --yes ppa:deadsnakes/ppa
+    - apt-get update --quiet
+    - apt-get install --assume-yes --no-install-recommends curl jq make python${PYTHON_MAJOR/2/}-pip python${PYTHON_VERSION}-dev
+    - python${PYTHON_VERSION} -m pip install --upgrade pip==10.0.1
+    - python${PYTHON_VERSION} -m pip install virtualenv==16.0.0
+    - rm -f /usr/bin/python; ln -s /usr/bin/python${PYTHON_VERSION} /usr/bin/python
+    - rm -f /usr/bin/pip; ln -s /usr/bin/pip${PYTHON_MAJOR} /usr/bin/pip
+    - virtualenv ~/venv
+    - source ~/venv/bin/activate
+    - pip install -r requirements-dev.txt
+    - scripts/fetch_secret.sh gcp-credentials.json > gcp-credentials.json
+    - export GOOGLE_APPLICATION_CREDENTIALS=$(pwd -P)/gcp-credentials.json
+    - sed -i -e s/dss.data.humancellatlas.org/dss.$DSS_TEST_STAGE.data.humancellatlas.org/ hca/default_config.json
+    - make install
+  script:
+    - make test
+  except:
+    - tags
+
+after_script:
+    - bash <(curl -s https://codecov.io/bash)
+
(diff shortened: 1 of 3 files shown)


scripts/fetch_secret.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -euo pipefail
+
+if [[ $# != 1 ]]; then
+	echo "This script fetches a secret value from AWS Secretsmanager, given a secret name."
```
xbrianh: tab carried over during copy/paste

.gitlab-ci.yml
```diff
@@ -0,0 +1,70 @@
...
+    - rm -f /usr/bin/pip; ln -s /usr/bin/pip${PYTHON_MAJOR} /usr/bin/pip
+    - virtualenv ~/venv
+    - source ~/venv/bin/activate
+    - pip install -r requirements-dev.txt
+    - scripts/fetch_secret.sh gcp-credentials.json > gcp-credentials.json
+    - export GOOGLE_APPLICATION_CREDENTIALS=$(pwd -P)/gcp-credentials.json
+    - sed -i -e s/dss.data.humancellatlas.org/dss.$DSS_TEST_STAGE.data.humancellatlas.org/ hca/default_config.json
+    - make install
+  script:
+    - make test
```
xbrianh: These two lines appear redundant with the job definitions.

mikebaumann: Thanks @xbrianh! Yes, these lines were redundant and have been removed
from the concrete job definitions.

FYI, this is a historical artifact of the process I used to develop the
GitLab CI configuration.
Most of the configuration development was done locally on my system,
using:
`$ gitlab-runner exec docker test-3.6`
Yet, using gitlab-runner in this way has some known limitations, documented here: https://docs.gitlab.com/runner/commands/#limitations-of-gitlab-runner-exec
Although not included in the documentation, it seems like use of `extends` is one of them.
The updated configuration works fine with the GitLab deployment, yet run locally results in:
```
$ gitlab-runner --debug exec docker test-3.6
Runtime platform                                    arch=amd64 os=darwin pid=3413 revision=4745a6f3 version=11.8.0
FATAL: missing 'script' for job                    
```
The redundant lines were part of the workaround for this during development, and are cleaned-up now.

scripts/fetch_secret.sh
mikebaumann: Fixed, thanks!

.gitlab-ci.yml
xbrianh: ah, that makes sense, thanks for the explanation :)

.gitlab-ci.yml
```diff
@@ -0,0 +1,58 @@
...
+  except:
+    - tags
+
+after_script:
+    - bash <(curl -s https://codecov.io/bash)
```
mikebaumann: Hi @xbrianh, I suspect this `after_script` section for code coverage can be removed, do you happen to know if that is true? In short, it appears the necessary code coverage instrumentation is in the Makefile.
Here are observations regarding why it is currently here, and why I suspect it can be removed:
* Looking at the `data-store` repo:
    * The `.travis.yml` has an entry like the above, yet the `.gitlab-ci.yml` does not.
    * The Makefile includes code coverage instrumentation and report consolidation
    * The GitHub PR checks include the codecov results
* Looking at the `dcp-cli`:
   * The `.travis.yml` has an entry like the above, the `.gitlab-ci.yml` currently does, yet perhaps *should not*
    * The Makefile includes code coverage instrumentation and report consolidation
    * The GitHub PR checks include the codecov results
So, is there any reason to keep the `after_script` section above?
I can try taking it out and checking for any untoward effects, but I'm not sure what all the potential undesired effects might be, so I thought I'd ask. :)

xbrianh: Your analysis indicates that we can drop `after_script` and expect code coverage to keep working as expected. I'm not aware of any possible negative affects, so I think we should give it a shot.

mikebaumann:
> The python versioning looks like it was a pain, congrats on that!
Thanks @DailyDreaming, yes, it was a bit of a pain -- thanks for your help with getting it worked out!
This was why my initial inclination was to use the Docker Hub official Python images (on Debian) with the given version of Python preinstalled. However, I agree Ubuntu makes a better base for debugging, etc. and is probably closer to what our end users on Linux are using, so a good choice for CLI testing. Thanks for the recommendation!

.gitlab-ci.yml
mikebaumann: Here's a description of what this does:

[Codecov Bash Uploader](https://docs.codecov.io/docs/about-the-codecov-bash-uploader)

Given that the PR code coverage reports are accessed at `https://codecov.io/...` this may be important, unless GitLab is providing the coverage reports some other way.
GitLab can publish coverage reports as described here:

[Publish code coverage report with GitLab Pages](https://about.gitlab.com/2016/11/03/publish-code-coverage-report-with-gitlab-pages/)

At present, it seems the code coverage reports on our GitHub PRs are coming from the Travis builds, which are running the CodeCove bash uploader, and are not coming from our GitLab builds. *If* that is the intended behavior, now and going forward, then I we we wouldn't need to, and shouldn't, run the Codecov Bash uploader for our GitLab builds. 
Correct? Comments?

PR #280, opened 2019-03-19
allow for not providing slashes when selecting upload area

parthshahva:
This is for ticket https://github.com/HumanCellAtlas/upload-service/issues/326. This allows users to pass in upload area uri without a slash when selecting it via hca cli

### DIFF
--- a/test/integration/upload/cli/test_select.py
+++ b/test/integration/upload/cli/test_select.py
@@ -22,18 +22,29 @@ class TestUploadCliSelectCommand(UploadTestCase):
 
     def setUp(self):
         super(self.__class__, self).setUp()
-        self.area_uuid = str(uuid.uuid4())
-        self.uri = "s3://org-humancellatlas-upload-test/{}/".format(self.area_uuid)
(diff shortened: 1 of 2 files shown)


test/integration/upload/cli/test_select.py
```diff
@@ -35,6 +35,17 @@ def test_when_given_an_unrecognized_urn_it_stores_it_in_upload_area_list_and_set
         self.assertEqual(self.uri, config.upload.areas[self.area_uuid]['uri'])
         self.assertEqual(self.area_uuid, config.upload.current_area)
 
+    def test_when_given_an_unrecognized_uri_without_slash_it_sets_it_as_current_area(self):
+        uri_without_slash = "s3://org-humancellatlas-upload-test/{}".format(self.area_uuid)
```
maniarathi: Instance variables are generally preceded with an underscore [0]. If you're feeling gracious, could make the change throughout the class.

[0] https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

PR #282, opened 2019-03-20
Updating boto3 version

maniarathi:
Previous 1.7-1.8 requirement is incompatible with upload service. Also alphabetized the dependencies.


xbrianh:
@parthshahva Is Upload prepared for this boto3 version bump?

sampierson:
@xbrianh yes Upload is the one initiating this

parthshahva:
@maniarathi @sampierson The upgrade of boto3 breaks the use of mock in our tests, leading to a slate of permission denied errors on CreateBucket. We'll need to fix our tests before merging this in.

kislyuk:
Only direct package dependencies (those imported by the library) should be listed here. Dependencies should manage their sub-dependencies.

PR #286, opened 2019-03-26
Windows Allspark Testing

amarjandu:
Add Windows env testing to Allspark
Resolves: #260


.gitlab-ci.yml
```diff
@@ -34,6 +34,16 @@ stages:
   except:
     - tags
 
+.windows-test-base:
+  stage: test
+  before_script:
```
amarjandu: add comment to note which version is being used by the runner system

.gitlab-ci.yml
```diff
@@ -34,6 +34,16 @@ stages:
   except:
```
amarjandu: cleanup linux section -- remove the make install for the base

Makefile
```diff
@@ -45,6 +46,21 @@ clean:
...
+
+clean-win:
+	cmd.exe /c "rmdir /s /q build 2>nul & exit 0 "
+	cmd.exe /c "rmdir /s /q dist 2>nul & exit 0 "
+	cmd.exe /c "for %%f in (*.egg-info) do rmdir /s /q %%f 2>nul & exit 0 "
+
+install-win: clean-win build-win
+	cmd.exe /c "for %%f in (dist\*.whl) do pip install --upgrade %%f"
+
+test-win: install-win integrationtests unittests
```
amarjandu: add coverage here

scripts/fetch_secret.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python
+
+import os
+import sys
+import boto3
+import json
+import argparse
+
+
```
amarjandu: remove white spaces.

Makefile
amarjandu: add lint-win to install

scripts/fetch_secret.py
amarjandu: flake8 this file

amarjandu:
@jesse Brennan, I’ll add that in. 
It’s all one-team one-dream over here, thanks for catching it :)

Makefile
amarjandu: added coverage to end of test-win, added lint-win

.gitlab-ci.yml
amarjandu: removed - make install in before_script

.gitlab-ci.yml
```diff
@@ -53,3 +63,8 @@ test-3.6:
   extends: .linux-test-base
   variables:
     PYTHON_VERSION: "3.6"
+
+test-3.6-win:
+  extends: .windows-test-base
+  tags:
+    - windows
```
mikebaumann: Please add a new line to the end of the file.

.gitlab-ci.yml
```diff
@@ -28,12 +28,22 @@ stages:
...
-    - make install
   script:
     - make test
   except:
     - tags
 
+.windows-test-base:
+  stage: test
+  before_script:
+    #currently 3.6.8 is offered by the windows-runner
```
mikebaumann: I understand that we will have more explicit control of the python version(s) used for testing when we move to using Docker images with the GitLab Windows Runner Helper. In the interim, please at least include the actual Python version in the build output. Better still would be to verify that the Python version being used is the expected version (e.g. 3.6) and fail the build if it is not.

.gitlab-ci.yml
```diff
@@ -53,3 +63,8 @@ test-3.6:
   extends: .linux-test-base
   variables:
     PYTHON_VERSION: "3.6"
+
+test-3.6-win:
+  extends: .windows-test-base
+  tags:
```
mikebaumann: You could specify the PYTHON_VERSION in the windows test job (as is being done for the Linux test jobs), then in then .windows-test-base verify that the installed version matches the PYTHON_VERSION  value.

.gitlab-ci.yml
```diff
@@ -34,6 +34,16 @@ stages:
...
     - tags
 
+.windows-test-base:
+  stage: test
+  before_script:
+    - virtualenv venv3
+    - venv3\Scripts\activate
+    - pip install -r requirements-dev.txt
+    - .\scripts\fetch_secret.py --secret-name gcp-credentials.json > gcp-credentials.json
+    - $env:GOOGLE_APPLICATION_CREDENTIALS = (Convert-Path .\gcp-credentials.json)
```
mikebaumann: Can this simply be changed to the following?
```
    - $env:GOOGLE_APPLICATION_CREDENTIALS = ./gcp-credentials.json

```

.gitlab-ci.yml
amarjandu: resolved

.gitlab-ci.yml
amarjandu: i dont believe so, addin Convert-Path will expand out the path to the directory just like, `export GOOGLE_APPLICATION_CREDENTIALS=$(pwd -P)/gcp-credentials.json` does in the linux base.

amarjandu:
> Why won't this test pass on windows? What's the failure?

currently the issue is with how tempfiles are 'reopened' and the windows OS. the downloading mechanism will work correctly, but the way way in which the integration test uses temp directories to store the downloaded data for the test does not work.

Ive attempted to move the path, and change the object class, but no change to the error.

jessebrennan:
Ok. I have a windows machine I can bring in on Monday if you want and we
can try and debug it.

.gitlab-ci.yml
amarjandu: Implemented print + check

.gitlab-ci.yml
amarjandu: implemented

PR #288, opened 2019-03-30
Add ingest facing endpoints to hca upload client library (upload-service/#211)

sampierson:
Per Upload ticket #211: Add ingest facing endpoints to hca upload client library.

I also took this opportunity to improve the Upload API Library, changing it from a few functions in the hca.upload module to a class based system (a lá GitHub, ZenHub etc).  The root of the Upload API Library is now the `UploadService` class, from which all else follows.

This series of commits has the following major results:
1. Flesh out the low level `UploadApi` class so it addresses all REST endpoints.
2. Expose those low-level endpoints via higher-level abstractions in `UploadArea`.
3. Move cache-of-upload-areas functionality out of UploadArea, use directly from `UploadConfig`.
4. Cap everything off with `UploadService` - a starting point for the API library that mints UploadAreas and gives access to the config.


kozbo:
what is this about

sampierson:
All will be revealed in due course.  Ignore this for the time being.
I'm currently just using this PR to invoke CI builds to see what results my changes get in Travis.

hca/upload/api_client.py
```diff
@@ -1,81 +1,238 @@
...
-                "PUT {url} returned {status}, {content}".format(
-                    url=url,
-                    status=response.status_code,
-                    content=response.content))
+    ApiClient is not normally called directly, it is used by UploadArea.
+    """
+
+    def __init__(self, deployment_stage, authentication_token=None):
+        self.deployment_stage = deployment_stage
+        # self.api_url_base = self._api_url(deployment_stage=deployment_stage)
```
parthshahva: Can this be removed if its no longer used?

hca/upload/api_client.py
```diff
@@ -1,81 +1,238 @@
...
+        :return: a dict containing an AWS AccessKey, SecretKey and SessionToken
+        :rtype: dict
+        :raises UploadApiException: if credentials could not be obtained
+        """
+        response = self._make_request("post", path="/area/{uuid}/credentials".format(uuid=area_uuid))
+        return response.json()
+
+    # Files
+
+    def store_file(self, area_uuid, filename, file_content, content_type):
```
parthshahva: Should this have a retry?

hca/upload/api_client.py
```diff
@@ -1,81 +1,238 @@
...
 
 
+class UploadApiException(RuntimeError):
+    """
+    Exception returned if problems are encountered
+    """
+    pass
+
+
 class ApiClient:
```
parthshahva: Should all the methods in this class be wrapped in retries?

hca/upload/api_client.py
sampierson: Many things in this library should probably have retries.  I think we should figure out which ones, and implement the retries, under a separate ticket.  I think there is enough in this one already.

hca/upload/api_client.py
sampierson: Ditto.

PR #292, opened 2019-04-05
Catch exception if link exists

jessebrennan:
This was causing tests to fail for datastore (Sorry :disappointed:)


jessebrennan:
@xbrianh I'm going to fix coverage, then I think we're good to go.

Issue #299, opened 2019-04-12
PUT collection should not require a version.


DailyDreaming:
The description says `If this is not provided, the version is automatically generated.` but this is a lie upon testing.
```
(v3nv) quokka@qcore:~/data-store$ hca dss put-collection -h
usage: hca dss put-collection -h --replica {aws,gcp} --uuid UUID --version
                              VERSION --contents CONTENTS [CONTENTS ...]
                              --description DESCRIPTION --details DETAILS
                              --name NAME

Create a new collection. Collections are sets of links to files, bundles,
other collections, or fragments of JSON metadata files. Each entry in the
input set of links is checked for referential integrity (the link target must
exist in the replica referenced). Up to 1000 items can be referenced in a new
collection, or added or removed using `PATCH /collections`. New collections
are private to the authenticated user. Collection items are de-duplicated (if
an identical item is given multiple times, it will only be added once).
Collections are replicated across storage replicas similarly to files and
bundles.

Optional Arguments:
  -h, --help            show this help message and exit

Required Arguments:
  --replica {aws,gcp}   Replica to write to.
  --uuid UUID           A RFC4122-compliant ID for the collection.
  --version VERSION     Timestamp of collection creation in DSS_VERSION format
                        format. If this is not provided, the version is
                        automatically generated.
  --contents CONTENTS [CONTENTS ...]
                        A list of objects describing links to files, bundles,
                        other collections, and metadata fragments that are
                        part of the collection.
  --description DESCRIPTION
                        A long description of the collection, formatted in
                        Markdown.
  --details DETAILS     Supplementary JSON metadata for the collection.
  --name NAME           A short name identifying the collection.
```

PR #304, opened 2019-04-16
Create Version Method

amarjandu:
Adds command `hca dss create-version` to print out a version stamp
to the user

Resolves: #299


amarjandu:
> ```
> (v3nv) quokka@qcore:~/dld/dcp-cli$ hca dss create-version -h
> usage: hca dss create-version -h
> 
> Print a timestamp that can be used for versioning
> 
> Optional Arguments:
>   -h, --help  show this help message and exit
> ```
> 
> Looks good! Maybe add an example too?


I cant find a way to cleanly insert a block of test showing the Example....

PR #309, opened 2019-04-19
Confirm hard link location during download (fixes #298)

### LINKED ISSUE #298: Confirm hard link location during download
When downloading a file with the bundle oriented download a link is made into the internal filestore. 

Currently this link isn't made if a link to the file already exists. This could be wrong in the case where an incorrect file with the same name already exists in the directory.

In the case where a link already exists, the correct behavior is to check that the source and destination link to the same inode.

jessebrennan:
Files in the download directory should only link to the filestore. If a link already exists this change ensure that it links to the filestore and raises an error if it links somewhere else.

### DIFF
--- a/test/unit/test_dss_client.py
+++ b/test/unit/test_dss_client.py
@@ -337,6 +337,16 @@ class TestManifestDownloadBundle(AbstractTestDSSClient):
     def test_download_dir_dot_dir(self):
         self._test_download_dir(os.path.join('.', 'a_nested_dir'))
 
+    @patch('hca.dss.DSSClient.get_bundle', side_effect=_fake_get_bundle)
+    @patch('hca.dss.DSSClient._download_file', side_effect=_fake_download_file)
+    def test_manifest_download_bad_file(self, _, __):
+        """
+        Ensure error is raised if a user created file has the same name as the one
+        we're trying to download.
+        """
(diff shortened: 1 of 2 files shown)


test/unit/test_dss_client.py
```diff
@@ -279,6 +279,17 @@ def test_manifest_download_bundle(self, _, __):
...
+
+    @patch('hca.dss.DSSClient.get_bundle', side_effect=_fake_get_bundle)
+    @patch('hca.dss.DSSClient._download_file', side_effect=_fake_download_file)
+    def test_manifest_download_bad_file(self, _, __):
+        """
+        Ensure error is raised if a user created file has the same name as the one
+        we're trying to download.
+        """
+        _touch_file(os.path.join(self.manifest[1][0], self.manifest[1][3]))
+        self.assertRaises(RuntimeError, self.dss.download_manifest, self.manifest_file, 'aws')
```
hannes-ucsc: Might be nice to have a bit more evidence that the root exception was indeed the IOError.

What I've done in the past, is to either `raise RuntimeError('Bla', list_of_underlying_exceptions[0:10])` and in the assertion examine the `args` attribute. Note 

```
>>> RuntimeError('bla', [IOError()]).args
('bla', [OSError()])
```

jessebrennan: @hannes-ucsc I'm not sure I understand what you're suggesting here.

PR #317, opened 2019-04-24
Add `--layout` option to manifest download

jessebrennan:


jessebrennan:
The diff is large but the actual substantive changes are small. Notice the two separate commits.

PR #323, opened 2019-04-30
Remove crcmod from direct dependencies

kislyuk:
It is no longer in use.

- Eliminates a binary dependency with no Linux wheels available on PyPI
- The binary extension build process introduces non-reproducibility into the build
- Developers should use crc32c instead where needed


sampierson:
Added @maniarathi as a  reviewer.  This module should have been removed when its use was replaced by crc32c.

Issue #296, opened 2019-04-11
Improve parallelism for bundle download


jessebrennan:
Currently all of the download tasks for a bundle are fetched in serial but the download is done in parallel.

This could be changed so that files are downloaded while other files are being fetched.

hannes-ucsc:
Ideally, this would involve only one thread pool.

hannes-ucsc:
Demo-ed on 5/13/2019

PR #325, opened 2019-04-30
Improve bundle parallelism (#296)

jessebrennan:
Note that this PR is based of off #317 and so should be merged after.

Now all of the information for each bundle is downloaded in parallel, and then all of the files for all of the undles are downloaded in parallel. Unavoidably both of these steps happen in sequence.

### DIFF
--- a/hca/dss/__init__.py
+++ b/hca/dss/__init__.py
@@ -7,6 +7,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import errno
 import functools
+import itertools
 from collections import defaultdict, namedtuple
 import csv
 import concurrent.futures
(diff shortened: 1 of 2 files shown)


hca/dss/__init__.py
```diff
@@ -361,15 +372,19 @@ def _download_manifest_bundle(self,
...
+                               download_dir="",
+                               metadata_files=('*',),
+                               data_files=('*',),
+                               num_retries=10,
+                               min_delay_seconds=0.25):
+        """
+        Returns an iterator of tasks that each download one of the files in a bundle.
+        """
+        logger.info('Downloading bundle %s version %s ...', bundle_uuid, version)
         bundle = self.get_bundle(uuid=bundle_uuid, replica=replica, version=version if version else None)["bundle"]
```
hannes-ucsc: As we noticed today, we need to do `get_bundle.iterate(…)` here to get all files, not just the first 500. This API change is insidious in that way because it silently truncates the response to 500 files.

hca/dss/__init__.py
```diff
@@ -361,20 +372,25 @@ def _download_manifest_bundle(self,
...
+                               download_dir="",
+                               metadata_files=('*',),
+                               data_files=('*',),
+                               num_retries=10,
+                               min_delay_seconds=0.25):
+        """
+        Returns an iterator of tasks that each download one of the files in a bundle.
+        """
+        logger.info('Downloading bundle %s version %s ...', bundle_uuid, version)
+        bundle = self.get_bundle.iterate(uuid=bundle_uuid, replica=replica, version=version if version else None)
```
hannes-ucsc: Switch to `.paginate()` which was added yesterday. Sorry for not alerting you earlier.

PR #335, opened 2019-05-07
Add paginate method.

DailyDreaming:
Added tests, since pagination/iteration should have some basic tests.

### DIFF
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -237,7 +237,7 @@ class _PaginatingClientMethodFactory(_ClientMethodFactory):
         For example, GET /bundles/{id} and GET /collections/{id} yield the
         items contained within; POST /search yields search result items.
         """
-        for page in _get_raw_pages(**kwargs):
+        for page in self._get_raw_pages(**kwargs):
             if page.json().get('results'):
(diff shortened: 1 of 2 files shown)


hca/util/__init__.py
```diff
@@ -225,6 +225,12 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
 class _PaginatingClientMethodFactory(_ClientMethodFactory):
     def iterate(self, **kwargs):
+        """
+        Yield specific items from each response depending on its contents.
+
+        For example, bundles and collections yield the files contained within and
+        a POST search yields result items.
```
kislyuk: Consider:

"For example, `GET /bundles/{id}` and `GET /collections/{id}` yield the items contained within; `POST /search` yields search result items."

hca/util/__init__.py
```diff
@@ -238,6 +244,13 @@ def iterate(self, **kwargs):
                 for collection in page.json().get('collections'):
                     yield collection
 
+    def paginate(self, **kwargs):
+        """Yield entire paged responses."""
```
kislyuk: Consider: "Yield paginated responses one response body at a time."

PR #342, opened 2019-05-16
Logout application_secret flush

amarjandu:
Remove application_secrets from config file during a logout.
Addresses issue if auth_uri has been updated from dss `/internal/application_secret`

### DIFF
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -381,11 +381,11 @@ class SwaggerClient(object):
         """
         Clear {prog} authentication credentials previously configured with ``{prog} login``.
         """
-        try:
-            del self.config["application_secrets"]
-            del self.config["oauth2_token"]
-        except KeyError:
-            pass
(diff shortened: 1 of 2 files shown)


hca/util/__init__.py
```diff
@@ -382,6 +382,7 @@ def logout(self):
         Clear {prog} authentication credentials previously configured with ``{prog} login``.
         """
         try:
+            del self.config["application_secrets"]
```
natanlao: If this line results in a KeyError, will `del self.config["oauth2_token"]` still run?

amarjandu: good catch, updated

amarjandu:
@natanlao  when we are testing login/logout to get access with the updated auth_uri in the DSS, we have to manually manipulate config files for the user, its easier for them to just logout, then login and all that information is refetched from the DSS for them..

PR #351, opened 2019-05-30
Add support for downloading collections

### LINKED ISSUE #339: Allow download of a complete Collection from the cli
## User story
* As an end-user of the Collections service 
* I want to download all the files in a Collection with a single cli command
* so that I don't have to download one item a time.

## Definition of done
* Add command e.g. "hca dss download-collection --uuid <collection_uuid>" that downloads all files in a Collection.

Reasonable limitations to start:
* Only support basic object types like bundles and files
* limit the number of levels of nesting when collections include other collections.
* Include some form of loop detection to avoid collections that end eventually including themselves through nesting.

natanlao:
Preliminary support for `hca dss download-collection`.

Features:
* Takes advantage of parallelization in `DSSClient.download_manifest`
* Supports downloading nested collections and bundles
* Properly handles collections that contain themselves. (For what it's worth, I can't find a way to create this case in the DSS, since it prevents me from creating collections that contain collections that don't yet exist.)
* Will not download collections that contain bundles/collections that are nested too deeply (depth 4 by default, configurable)

Limitations:
* Files specified in a collection independent of a bundle cannot be downloaded. This is a limitation of `DSSClient.download_manifest`.

Closes #339

### DIFF
--- a/hca/dss/__init__.py
+++ b/hca/dss/__init__.py
@@ -691,6 +691,7 @@ class DSSClient(SwaggerClient):
                 if (obj['uuid'], obj['version']) in _ignore:
                     logger.info("Ignoring already-seen collection %s version %s",
                                 obj['uuid'], obj['version'])
+                    continue
                 _ignore.append((obj['uuid'], obj['version']))
                 r = self._serialize_col_to_manifest(uuid=obj['uuid'],
                                                     replica=replica,
--- a/hca/dss/__init__.py
+++ b/hca/dss/__init__.py
@@ -722,8 +722,6 @@ class DSSClient(SwaggerClient):
             download the latest. The version is a timestamp of bundle creation
             in RFC3339
         :param str download_dir: The directory into which to download
-        :param int max_depth: The amount of "levels" of the collection to
-            tolerate before raising an error
 
         Download a bundle and save it to the local filesystem as a directory.
         """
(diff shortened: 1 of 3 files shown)


hca/dss/__init__.py
```diff
@@ -610,3 +611,104 @@ def _parse_manifest(cls, manifest):
...
+            used to recognize collections that eventually include themselves
+        """
+        if _max_depth < 1:
+            raise RuntimeError("Maximum depth reached")
+        errors = 0
+        rows = []
+        for obj in self.get_collection(uuid=uuid, replica=replica, version=version)['contents']:
+            if obj['type'] == 'file':
+                # Currently cannot download files not associated with a
+                # bundle. This is a limitation of :meth:`download_manifest`
```
natanlao: I've opened HumanCellAtlas/dcp-cli#356 to address this.

test/unit/test_dss_client.py
```diff
@@ -7,13 +7,18 @@
...
 
 import six
 from mock import patch
 
-from hca.util.compat import walk
+from hca.util.compat import USING_PYTHON2, walk
 from hca.dss import DSSClient
 
+if USING_PYTHON2:
+    import backports.tempfile
```
DailyDreaming: Could be `import backports.tempfile as tempfile`.

hca/dss/__init__.py
```diff
@@ -610,3 +611,107 @@ def _parse_manifest(cls, manifest):
...
+        serialize the collection into a set of dicts that that can be
+        used to generate a manifest file.
+
+        Most of the heavy lifting is handled by
+        :meth:`DSSClient.download_manifest`.
+
+        :param uuid: uuid of the collection to serialize
+        :param replica: replica to query against
+        :param version: version of the specified collection
+        :param int _max_depth: maximum depth to serialize - used to limit
```
DailyDreaming: Depth shouldn't matter or be a limiting factor here.  Consider using a while loop instead of recursion.

Issue #112, opened 2018-04-26
hca prints --help insted of run


oskotsky:
hca dss download --bundle-uuid XXXXXXXXXX --replica aws just prints help output. adding --log-level DEBUG is not recognized as a switch 

Thank you

ttung:
What do you get when you run `hca --version`?

oskotsky:
$ hca --version
hca 3.4.6 (CPython 3.4.2 Linux-3.10.0-862.el7.x86_64-x86_64-with-redhat-7.5-Maipo)

I am trying to run this on rhel7 remote server, no gui. Would it be a problem? Does hca use port 443 only? I wonder if any additional traffic blocked by our firewall and that is why it breaks. 
Thank you

ttung:
What's the exact command you're running?

oskotsky:
I was given a script "hca-download-census". Inside the script this seems to fail: "hca dss download --bundle-uuid XXXXXXXXXX --replica aws". I tried on my home system - Ubuntu Desktop and it worked without any issues.

ttung:
That looks good.  Can you zip up the contents of `~/.config/hca` and post that here?

oskotsky:
[hca_config.zip](https://github.com/HumanCellAtlas/dcp-cli/files/1953407/hca_config.zip)
here it is

Thank you

ttung:
Hm... that looks okay.  What do you get when you run `hca dss download --help`?

oskotsky:
i see this:

$ hca dss download --help
usage: hca dss download [-h] --bundle-uuid BUNDLE_UUID --replica REPLICA
                        [--version VERSION] [--dest-name DEST_NAME]

Download a bundle and save it to the local filesystem as a directory.

optional arguments:
  -h, --help            show this help message and exit
  --bundle-uuid BUNDLE_UUID
  --replica REPLICA
  --version VERSION
  --dest-name DEST_NAME

ttung:
So what do you get when you run:

`hca dss download --bundle-uuid c94a43f9-257f-4cd0-b2fe-eaf6d5d37d18 --replica aws`?

oskotsky:
this is what is odd:

$ hca dss download --bundle-uuid c94a43f9-257f-4cd0-b2fe-eaf6d5d37d18 --replica aws
usage: hca dss [-h]
[...]

PR #359, opened 2019-06-13
Begin DCP Query Service CLI

kislyuk:
Because the DCP Query Service uses Swagger 3.0/OpenAPI 3.0, this also begins OpenAPI 3.0 compatibility work in the Swagger Client.

Closes https://github.com/HumanCellAtlas/query-service/issues/185

### DIFF
--- a/hca/query/cli.py
+++ b/hca/query/cli.py
@@ -1,0 +1,17 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import sys
+
+from . import DCPQueryClient
+
+def add_commands(subparsers, help_menu=False):
+    query_parser = subparsers.add_parser('query', help="Interact with the HCA DCP Query Service")
+
+    def help(args):
+        query_parser.print_help()
+
+    if sys.version_info >= (2, 7, 9):  # See https://bugs.python.org/issue9351
+        query_parser.set_defaults(entry_point=help)
+    query_subparsers = query_parser.add_subparsers()
+    query_cli_client = DCPQueryClient()
+    query_cli_client.build_argparse_subparsers(query_subparsers, help_menu=help_menu)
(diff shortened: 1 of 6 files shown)


hca/query/cli.py
```diff
@@ -0,0 +1,17 @@
...
+
+from . import DCPQueryClient
+
+def add_commands(subparsers, help_menu=False):
+    query_parser = subparsers.add_parser('query', help="Interact with the HCA DCP Query Service")
+
+    def help(args):
+        query_parser.print_help()
+
+    if sys.version_info >= (2, 7, 9):  # See https://bugs.python.org/issue9351
```
DailyDreaming: Do we need to worry about python < `2.7.9`?

hca/util/__init__.py
```diff
@@ -283,9 +283,14 @@ def __init__(self, config=None, swagger_url=None, **session_kwargs):
         self.methods = {}
         self.commands = [self.login, self.logout, self.refresh_swagger]
         self.http_paths = collections.defaultdict(dict)
-        self.host = "{scheme}://{host}{base}".format(scheme=self.scheme,
-                                                     host=self.swagger_spec["host"],
-                                                     base=self.swagger_spec["basePath"])
+        if "openapi" in self.swagger_spec:
```
DailyDreaming: Nit: Would it be possible to have a variable similar to `USING_PYTHON2` specifying the swagger version?

hca/query/cli.py
kislyuk: See https://github.com/HumanCellAtlas/dcp-cli/issues/112

hca/util/__init__.py
kislyuk: I don't think hardcoding the version of the API spec language in each client is necessary/good, because it makes upgrading the server-side version harder. Instead the client should ideally detect the version of the server-supplied spec and adjust accordingly (which is what is done here).

I do agree that version-specific behavior could be better separated in this class. In the interest of expediency I just put a few if statements here, but ideally we would separate the code more thoroughly or even factor it out into an "OpenAPISpecParser" and "SwaggerAPISpecParser" or something.

kislyuk:
Ignoring codeclimate report. Also, the coverage reporting seems to be broken.

Issue #231, opened 2019-02-14
Download bundle manifest as bundle.json


diekhans:
`hca dss download` and `hca dss download-manifest` do not include `bundle.json`
files in the download.  The `bundle.json` contains information not available in any of the downloaded metadata files and is key to understanding the bundle.

hannes-ucsc:
To clarify, `bundle.json` is an arbitrary name that bears no significance in the metadata or the DSS. The DSS refers to this JSON structure as the *bundle manifest*. It *is* the object that represents a bundle in the DSS and that makes up the response of `GET /bundles/$uuid`.

It does make sense to store the bundle manifest in a file called `bundle.json` on the user's disk. To fit into the `.hca` cache implemented in #234, the primary location of this file should be `.hca/bundles/$bundle_fqid` with a user-friendly hardlink in `$bundle_fqid/bundle.json`

theathorn:
@jessebrennan to evaluate potential bug with implementation and raise isssue.

theathorn:
@jahilton  automation is not required for General Availability. Removing from epic.

jessebrennan:
This needs to be released.

PR #361, opened 2019-07-10
Download bundle metadata into bundle.json (#231)

jessebrennan:


hca/dss/__init__.py
```diff
@@ -504,6 +534,15 @@ def _download_file(self, dss_file, dest_path, num_retries=10, min_delay_seconds=
                 raise ValueError("Expected sha256 {} Received sha256 {}".format(
                     dss_file.sha256.lower(), download_hash.lower()))
 
+    def _make_path_if_necessary(self, dest_path):
+        directory, _ = os.path.split(dest_path)
+        if directory:
+            try:
+                os.makedirs(directory)
```
GPelayo: Instead of using the try-catch. Do...
```suggestion
            os.makedirs(directory, exist_ok=True)
```

GPelayo: Nevermind the `exist_ok` keyword arg in `os.makedirs()` doesn't exist in python2

PR #393, opened 2019-07-17
Skip download-collections tests on Windows

natanlao:
`hca dss download-collections` relies on `hca.dss.download_manifest`, which [currently is not tested on Windows](https://github.com/HumanCellAtlas/dcp-cli/blob/7847a6a0c48d38777bda2740843cac23bfef2bd5/test/unit/test_dss_client.py#L187), causing Windows tests for #351 to fail. (I didn't catch that, sorry...)

This PR skips `download-collections` unit tests in the same way that `download_manifest` unit tests are skipped.

### DIFF
--- a/test/integration/dss/test_dss_cli.py
+++ b/test/integration/dss/test_dss_cli.py
@@ -165,6 +165,7 @@ class TestDssCLI(unittest.TestCase):
             # Deleting bundles is a privilege, not a right
             assert e.code == 403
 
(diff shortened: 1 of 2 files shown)


natanlao:
(And [confirmed passing](https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/dcp-cli/pipelines/13176) on Allspark.)

PR #394, opened 2019-07-17
Add an option to delete the cache when download completes.

DailyDreaming:
Adds functionality to address: https://github.com/HumanCellAtlas/data-store/issues/2064

In the future, we might consider setting the default to `True` for this, but in the meantime, this should be a useful switch for the needs of the `dss`.

Will add a test if it makes sense to add this.


hca/dss/__init__.py
```diff
@@ -610,6 +625,10 @@ def _do_download_file(self, dss_file, fh, num_retries, min_delay_seconds):
                 raise
         return hasher.hexdigest()
 
+    @classmethod
+    def _dir_path(cls, download_dir):
```
hannes-ucsc:
```suggestion
    def _filestore_path(cls, download_dir):
```

hca/dss/__init__.py
```diff
@@ -272,6 +281,9 @@ def download_manifest(self,
         :param float min_delay_seconds: The minimum number of seconds to wait in between retries for downloading any
             file
         :param str download_dir: The directory into which to download
+        :param bool delete_cache: When downloading files, the folder '.hca' contains duplicate hardlinks that serve as
```
hannes-ucsc:
```suggestion
        :param bool delete_cache: When downloading files, the folder '.hca' contains additional file references (hard links) that serve as
```

hca/dss/__init__.py
```diff
@@ -272,6 +281,9 @@ def download_manifest(self,
...
+                                  a cache when downloading.  Specifying this option will delete that cache after the
+                                  files are downloaded.
```
hannes-ucsc: Add: 

> Note that deleting the file store will prevent you from efficiently switching the download layout (see --layout) or rerunning the download, potentially with additional rows in the manifest. Once you delete the file store, the next invocation will download all files again. Also note that the cache does not take up significant disk space. Each file is be stored exactly once, the file store contains only references (hard links) to the downloaded files.

hca/dss/__init__.py
```diff
@@ -216,6 +218,9 @@ def download(self,
         :param int num_retries: The initial quota of download failures to accept before exiting due to
             failures. The number of retries increase and decrease as file chucks succeed and fail.
         :param float min_delay_seconds: The minimum number of seconds to wait in between retries.
+        :param bool delete_cache: When downloading files, the folder '.hca' contains duplicate hardlinks that serve as
```
hannes-ucsc:
```suggestion
        :param bool delete_cache: When downloading files, the folder '.hca' contains additional file references (hard links) that serve as
```

hca/dss/__init__.py
```diff
@@ -216,6 +218,9 @@ def download(self,
...
+                                  a cache when downloading.  Specifying this option will delete that cache after the
+                                  files are downloaded.
```
hannes-ucsc: Add: 

> Note that deleting the file store will prevent you from efficiently downloading another bundle that shares files with a previously downloaded one. Once you delete the file store, the next invocation will download all files again. Also note that the cache does not take up significant disk space. Each file is be stored exactly once, the file store contains only references (hard links) to the downloaded files.

hannes-ucsc (changes requested):
I will probably oppose ever making this default to True. My review comments add language that highlights the benefit of the file store  and the implications of removing it.

I suggest that we decide if we want to use the term "cache" or "file store". I actually prefer "cache" but you would have to eliminate all mention of "file store" from both the code and the documentation. I think sticking with "file store" is easier. Whatever we do, we cannot afford to use inconsistent language.

DailyDreaming:
@hannes-ucsc All comments should be addressed.

jessebrennan:
@DailyDreaming are you still making changes or is this good to review?

jessebrennan (changes requested):
I know there's been mixed opinions about this, but I'm starting to feel more strongly that we should use the name filestore instead of cache. Here's my reasoning:

For a cache, you typically expect that it can be cleared / deleted and the only impact will be upon performance in the future. In the case of our filestore, deleting it can have dire consequences for some kinds of downloads.

Specifically, consider what would happen if the `download_manifest` command is run with the `delete_cache` option set and `layout=none`. The files would be downloaded to the filestore, the manifest would be updated with paths to the files, then all of the files would be deleted and the paths would be broken. After hours of downloading, the user ends up with a bunch of broken paths. This is definitely not ever desirable behavior. Using the name cache gives users the false impression that it can be safely deleted, or that running with these options could even make sense.

The example above has also made me question adding the `delete_cache` feature. I think (at the very least) the name should be changed to `delete_filestore`, and if the feature is maintained, it should turn off the writing of the updated manifest or even be removed completely from the manifest download. Also, if it is kept, it should come with a stronger warning, explaining that bundles downloaded with this option will use double the disk space for redundant files, etc.

Maybe we should with @hannes-ucsc, all talk about with together in person.

hannes-ucsc:
I hadn't thought about the case with `--layout=none`, that @jessebrennan brought up. `--layout=none` directly contradicts `--delete-cache` because using both would render a rewritten manifest with paths to non-existent files. 

We could of course revert the renaming, ensure that `--delete-filestore` can't be used in conjunction with `--layout=none` by making them mutually exclusive.

I have a third option: we enforce the specification of a download directory instead of polluting the current working directory. If the user is forced to chose the name of a directory for the download, the user will then be more likely to regard that entire directory as an artifact of the download and  therefore delete that directory instead of the file files in it. Deleting the directory would remove the hardlinks created for the selected layout, the rewritten manifest and the filestore.

jessebrennan:
@hannes-ucsc I think your third option is a good idea and perhaps we should make a separate ticket about making the download directory a required argument. It's worth pointing out however, that the use case @DailyDreaming was originally trying to solve is not addressed by that change.

PR #399, opened 2019-07-23
Batch Checkout for hca dss download [wip]

amarjandu:
Threads attempt to initiate checkouts before downloading data.
Does not force thread to stay with file as it attempts to 'retry' and get into the checkout bucket. 
Queue file downloads by size of the file, smaller to larger.

### DIFF
--- a/hca/dss/__init__.py
+++ b/hca/dss/__init__.py
@@ -351,7 +351,8 @@ class DSSClient(SwaggerClient):
         with concurrent.futures.ThreadPoolExecutor(self.threads) as executor:
             bundle_futures_to_uuid = {executor.submit(bundle_download_task): bundle_uuid
                                       for bundle_uuid, bundle_download_task in
-                                      self._download_manifest_tasks(manifest,
+                                      self._download_manifest_tasks(executor,
+                                                                    manifest,
                                                                     replica,
                                                                     num_retries,
                                                                     min_delay_seconds,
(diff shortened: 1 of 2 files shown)


hca/dss/__init__.py
```diff
@@ -417,15 +422,31 @@ def _bundle_download_tasks(self,
...
+    def _try_file_checkout(self, executor, download_dir, dss_file, file_path, num_retries, min_delay_seconds):
+        logger.info('Getting checkout status for file: %s', dss_file.name)
+        response = self.get_file._request(
+            dict(uuid=dss_file.uuid, version=dss_file.version, replica=dss_file.replica),
+            retries=False, stream=True)
+        logger.debug('status_code %s for file %s', response.status_code, dss_file.name)
+        if response.status_code == 301:
+            logger.info('file: %s still checking out...', dss_file.name)
+            executor.submit(self._try_file_checkout, executor, download_dir, dss_file, file_path, num_retries,
+                            min_delay_seconds)
```
xbrianh: Is the value of the Retry-After header honored somewhere in the download logic?

hca/dss/__init__.py
```diff
@@ -417,15 +422,31 @@ def _bundle_download_tasks(self,
...
+        elif response.status_code is 302 or 200:
```
natanlao: This should be

```suggestion
        elif response.status_code in (302, 200):
```

The current line is analogous to

```
    elif (response.status_code is 302) or 200:
```

and since `bool(200) == True`, this condition will always be met.

amarjandu:
Closing this PR; waiting for community review for presigned URLS.

PR #400, opened 2019-07-23
Clarity improvement to download-collection

natanlao:
This is better than what I had written previously. Thanks @amarjandu!


natanlao:
Passing on Allspark. https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/dcp-cli/pipelines/13464

PR #403, opened 2019-07-24
Fix bugs in bundle.json download

jessebrennan:
This PR fixes several things:
- Renames the internal variables to make it more clear that the `bundle.json` is the bundle manifest
- Fixes the bundle timestamp to be the bundle version
- Return the files in `bundle.json` as an array (instead of a dict like before)

### DIFF
--- a/hca/dss/__init__.py
+++ b/hca/dss/__init__.py
@@ -59,12 +59,12 @@ class DSSFile(namedtuple('DSSFile', ['name', 'uuid', 'version', 'sha256', 'size'
                    replica=replica)
 
     @classmethod
-    def from_bundle_json(cls, metadata_bytes, bundle_uuid, replica):
+    def from_bundle_json(cls, manifest_bytes, bundle_uuid, replica):
         return cls(name='bundle.json',
                    uuid=bundle_uuid,
                    version=datetime.utcnow().strftime("%Y-%m-%dT%H%M%S.%fZ"),
-                   sha256=hashlib.sha256(metadata_bytes).hexdigest(),
-                   size=len(metadata_bytes),
+                   sha256=hashlib.sha256(manifest_bytes).hexdigest(),
+                   size=len(manifest_bytes),
                    indexed=False,
                    replica=replica)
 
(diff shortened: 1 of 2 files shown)


hca/dss/__init__.py
```diff
@@ -426,25 +431,27 @@ def _bundle_download_tasks(self,
...
         self._make_path_if_necessary(file_path)
         hardlink(dest_path, file_path)
 
-    def _get_full_bundle_metadata(self, bundle_uuid, replica, version):
+    def _get_full_bundle_manifest(self, bundle_uuid, replica, version):
         """
         Takes care of paging through the bundle and checks for name collisions
         """
         pages = self.get_bundle.paginate(uuid=bundle_uuid, replica=replica, version=version if version else None)
         files = {}
```
amarjandu: is files used?

jessebrennan:
Yes, files is used to confirm that there are no names collisions within a
bundle which would cause download errors on MacOS.

hca/dss/__init__.py
```diff
@@ -59,10 +59,10 @@ def from_dss_bundle_response(cls, file_dict, replica):
                    replica=replica)
 
     @classmethod
-    def from_bundle_json(cls, manifest_bytes, bundle_uuid, replica):
+    def from_bundle_json(cls, manifest_bytes, bundle_uuid, version, replica):
         return cls(name='bundle.json',
```
jessebrennan: Explain why pretend bundle.json is a dssfile.

hca/dss/__init__.py
```diff
@@ -59,10 +59,10 @@ def from_dss_bundle_response(cls, file_dict, replica):
                    replica=replica)
 
     @classmethod
-    def from_bundle_json(cls, manifest_bytes, bundle_uuid, replica):
+    def from_bundle_json(cls, manifest_bytes, bundle_uuid, version, replica):
```
jessebrennan: Rename to `for_bundle_manifest`

jessebrennan:
@hannes-ucsc also approved this but in person, overlooking my computer.

Issue #412, opened 2019-07-31
Drop support for Python 3.4


sampierson:
Which was EOLd 2019-03-18

PR #413, opened 2019-07-31
[easy] Remove python3.4 (EOL) testing.

DailyDreaming:
Addresses #412 .


sampierson:
Python 3.4 is still listed in `setup.py`.   If we're not testing it against a release, shouldn't we stop declaring support for that release?

Issue #406, opened 2019-07-24
Refactor task scheduling for file downloads


hannes-ucsc:
The work for #405 requires tasks to submit follow-on tasks. The current approach of first enumerating all tasks and then submitting them does not make that easy.  The current apporoach would also benefit from the ad-hoc creation of tasks.

Refactor the current approach to make the thread pool executor and list of outstanding futures available to every method in the class. Consider making those two objects part attributes of the client class. Consider extracting all download-related state in the client class into another class whose instances are scoped to one download rather than the entire client. Consider moving some of the download-related functionality into methods of that class.

PR #414, opened 2019-08-01
Refactor download (#406)

jessebrennan:
Refactor download components into DownloadContext. Also make wrapper for parallelization that simplifies adding tasks.


jessebrennan:
@hannes-ucsc, your suggested changes from our face to face review earlier have been implemented.

Issue #418, opened 2019-08-06
Version conflict when installing with `pip install -e .`


jessebrennan:
To recreate:
```
$ virtualenv .venv -p python3.6
$ source .venv/bin/activate
$ pip --version
pip 19.2.1 from /home/jesse/gi/dcp-cli/.venv/lib/python3.6/site-packages/pip (python 3.6)
# be in base dir of repository
$ pip install -e .
$ hca dss download-manifest --replica aws --manifest manifest.tsv
Traceback (most recent call last):
  File "/home/jesse/gi/dcp-cli/.venv/lib/python3.6/site-packages/pkg_resources/__init__.py", line 574, in _build_master
    ws.require(__requires__)
  File "/home/jesse/gi/dcp-cli/.venv/lib/python3.6/site-packages/pkg_resources/__init__.py", line 892, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/home/jesse/gi/dcp-cli/.venv/lib/python3.6/site-packages/pkg_resources/__init__.py", line 783, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (docutils 0.15.2 (/home/jesse/gi/dcp-cli/.venv/lib/python3.6/site-packages), Requirement.parse('docutils<0.15,>=0.10'), {'botocore', 'awscli'})

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/jesse/gi/dcp-cli/.venv/bin/hca", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/home/jesse/gi/dcp-cli/.venv/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3088, in <module>
    @_call_aside
  File "/home/jesse/gi/dcp-cli/.venv/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3072, in _call_aside
    f(*args, **kwargs)
  File "/home/jesse/gi/dcp-cli/.venv/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3101, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/home/jesse/gi/dcp-cli/.venv/lib/python3.6/site-packages/pkg_resources/__init__.py", line 576, in _build_master
    return cls._build_from_requirements(__requires__)
[...]

PR #419, opened 2019-08-06
Avoid version conflict by pinning docutils (#418)

jessebrennan:


kislyuk:
> I think the CLI should always pin dependencies.

The CLI is liable to be installed alongside other Python libraries, increasing the likelihood of unsatisfiable pins if another package also depends on a pinned version of the same package.

For this reason libraries should range their dependencies as loosely as reasonable beyond the major version number.

@jessebrennan can you please describe the version conflict that occurred here, or link to a relevant issue?

jessebrennan:
@kislyuk it's issue #418. I linked it to the PR and referenced it in the PR title and commit.

kislyuk:
Thanks, sorry I missed that. LGTM

hannes-ucsc:
> The CLI is liable to be installed alongside other Python libraries, increasing the likelihood of unsatisfiable pins if another package also depends on a pinned version of the same package.

> For this reason libraries should range their dependencies as loosely as reasonable beyond the major version number.

In my experience, this approach doesn't work in practice, simply because "as loosely as reasonable" is ill-defined and unreliable. Any package can potentially break on any update of a dependency, even minor ones. This problem is the reason virtual environments were invented, and it's the reason why other languages use the share-nothing approach to package management, most prominently Node, Go and Rust.

PR #426, opened 2019-08-13
Show errors with `hca dss upload -q`

### LINKED ISSUE #188: -q flag silences error messages
The `-q` flag for `hca upload files` means "quiet" and "Suppress normal output." Setting it silences all upload error messages, which is probably a little quieter than most users would expect.

It's a bit worse when directly calling `hca.upload.UploadArea.upload_files`. That has a `report_progress` kwarg that silences all upload errors when `False`, which is the default. You would think `report_progress` turns off a progress bar or something.

natanlao:
Closes #188.

This PR changes the behavior of `hca dss upload -q` to show error messages where they were previously suppressed. This fix revealed an apparent misconfiguration in test.integration.upload.cli.test_upload that caused some failing tests to be erroneously reported as passing.

### DIFF
--- a/test/integration/upload/cli/test_upload.py
+++ b/test/integration/upload/cli/test_upload.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # coding: utf-8
-
+import itertools
 import os
 import sys
 import unittest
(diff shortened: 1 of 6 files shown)


test/integration/upload/cli/test_upload.py
```diff
@@ -24,6 +25,19 @@ def setUp(self):
         self.area = self.mock_current_upload_area()
         self.test_files = ['LICENSE', 'README.rst']
 
+    def upload_command(self, args):
```
DailyDreaming: Nit:
```
        paths = []
        if args.target_filename:
            paths.append(args.target_filename)
        for arg in args.upload_paths:
            if os.path.isdir(arg):
                paths.extend(itertools.chain.from_iterable((f for _, _, f in os.walk(arg))))
            else:
                paths.append(arg)
        for path in paths:
            self.add_upload_mock(self.area.uuid, path)
        return UploadCommand(args)
```

Could be shortened to:

```
        for arg in args.upload_paths + [args.target_filename or '']:
            if os.path.isdir(arg):
                [self.add_upload_mock(self.area.uuid, p) for l in [f for _, _, f in os.walk(arg)] for p in l]
            elif os.path.isfile(arg):
                self.add_upload_mock(self.area.uuid, arg)
        return UploadCommand(args)
```

Haha, you can ignore that one though if you like.

Though I'm wondering why add these mocked responses since you're just introducing a logging message condition?  Were the tests not running before without them?

natanlao: > Could be shortened to:

Updated with a similar version!

> Were the tests not running before without them?

Yes - the previous error suppression mechanism didn't just suppress logging of errors, it prevented errors from being raised at all, which caused tests that shouldn't have passed to pass silently. (All of those tests are fixed in this PR, though.)

Issue #255, opened 2019-03-05
Review log verbosity produced by file download from manifest


theathorn:
From @matthewspeir:
- A verbosity setting to hide warning/retry or other unnecessary messages would be great. Let me decide how much info I want to see or hide the details and put them in a log file.
- Basically I just want to know if it succeeded in downloading a file (and sometimes I don’t even care to see that.)

Bento007:
```$ hca --help
usage: hca [-h] [--version] [--log-level {DEBUG,INFO,WARNING,CRITICAL,ERROR}]
           {help,upload,dss}
```
In practice it looks like
``hca --log-level INFO dss ...`

amarjandu:
## Notes 4/8/19 : 9:20 AM

### Complete
Tested `--log-level {DEBUG,INFO,WARNING}` commands to see if they operate correctly.
`--log-level` works as expected, setting level to `INFO` prints out only bundle information as requested.

PR #432, opened 2019-08-23
Add progress bar to `hca dss upload` (closes #117)

### LINKED ISSUE #117: cli progress bar
Hi,

If easy enough to add, it would be really nice to have a progress bar so when us human brokers are uploading lots of data we had an idea where we are and an idea of how long it is going to take. In the past I have used tqdm for this and found it really good and easy to wrap around loops. https://github.com/tqdm/tqdm

Thanks!

natanlao:
Adds a progress bar to `hca dss upload`. This PR introduces a new dependency, `tqdm`. (This dependency is installed by `twine`, which is specified in `requirements-dev.txt`.)

`tqdm` itself has no dependencies, so while I try to avoid adding anything to `requirements.txt`, this seems more or less harmless.

Test with `hca dss upload --replica aws --staging-bucket org-humancellatlas-dss-cli-test --src-dir some_directory_large_enough_to_take_time_uploading/ --progress`. `--progress` is disabled by default because it messes with some of the integration tests.


amarjandu:
@DailyDreaming @natanlao i think there might be a case to add a key like `verbose_progress` into the configuration at some point. not sure if everyone wants to update their scripts to remove the verbose output. 

Just something that has been brought up before https://github.com/HumanCellAtlas/dcp-cli/issues/255 
Or perhaps it can be set with `log_level` which is in the config already.

PR #438, opened 2019-09-09
Remove six as a dependency (closes #425)

### LINKED ISSUE #409: Drop support for Python 2.7
_“DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won’t be maintained after that date. A future version of pip will drop support for Python 2.7.“_

The Tech Arch team has approved dropping Python 2.7 support.

Please note that `dcplib`, on which dcp-cli depends, is dropping support for Python 2.7 after version 2.1.2.

### LINKED ISSUE #425: Make `--layout=bundle` the default
Originally brought up by @ESapenaVentura (on Slack, `#data-store`):

> I tried hca dss download-manifest  this morning with a modified manifest (deleted almost all lines because I just wanted to test how it function), and it downloaded the files to the temporary hidden folder and, in the end, it modified the manifest to indicate the hidden folders where the files (with modified names) are. I tried downloading by bundle id and the behaviour is completely different, downloads the bundle at a non-hidden directory with the original name of the file. Is the first behaviour intended?

### LINKED ISSUE #435: Delete "six" as a dependency.
We no longer support python 2.7, so this should be removed.  Depends on #430.

natanlao:
Closes #409 
Closes #435 
Related to #430

### DIFF
--- a/hca/upload/upload_area_uri.py
+++ b/hca/upload/upload_area_uri.py
@@ -1,4 +1,5 @@
-from six.moves.urllib.parse import urlparse
+# -*- coding: utf-8 -*-
+import urllib.parse
 
 
 class UploadAreaURI:
@@ -11,7 +12,7 @@ class UploadAreaURI:
 
     def __init__(self, uri):
         self.uri = uri
-        self.parsed = urlparse(self.uri)
+        self.parsed = urllib.parse.urlparse(self.uri)
 
     def __str__(self):
         return self.uri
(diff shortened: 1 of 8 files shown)


natanlao:
@jessebrennan Ack, good catch, thanks!

hca/upload/upload_area_uri.py
```diff
@@ -1,4 +1,5 @@
-from six.moves.urllib.parse import urlparse
+# -*- coding: utf-8 -*-
```
jessebrennan: Also, what's up with this like?

natanlao:
> Nice. When you merge, can you consolidate 60aa8f4 and 291f26c? (The other commit makes sense to keep separate.)

Yep! I usually wait until review has completed to squash so that it doesn't interfere with comments.

jessebrennan:
That's an excellent strategy.

PR #440, opened 2019-09-11
Fix broken import causing integration test failures

natanlao:
There is limited to no `pty` support on Windows, so test cases calling it are skipped on Windows. Turns out that even importing `pty` can cause bad. This PR changes test behavior to only import `pty` on non-NT platforms.

### DIFF
--- a/test/integration/dss/test_dss_cli.py
+++ b/test/integration/dss/test_dss_cli.py
@@ -4,7 +4,6 @@ import contextlib
 import filecmp
 import json
 import os
-import pty
 import sys
 import unittest
 import uuid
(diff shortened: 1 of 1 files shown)


natanlao:
https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/dcp-cli/commit/36b711c7813ae78a45a4e3a45367ce84bf56c379

test/integration/dss/test_dss_cli.py
```diff
@@ -131,6 +130,7 @@ def test_upload_progress_bar(self):
     @unittest.skipIf(os.name is 'nt', 'No pty support on Windows')
     def test_upload_progress_bar_interactive(self):
         """Tests upload progress bar with a simulated interactive session"""
+        import pty
```
jessebrennan: Consider adding a comment explaining why this import is in an unusual place.

natanlao: Done.

PR #446, opened 2019-09-18
Refresh session token when the jwt expires.

DailyDreaming:


hca/util/__init__.py
```diff
@@ -131,16 +142,15 @@ def _request(self, req_args, url=None, stream=False, headers=None):
...
-            session = self.client.get_authenticated_session()
-        else:
-            session = self.client.get_session()
+        res = self._make_authenticated_request(body, headers, url, query, stream)
+
+        keep_session_alive = True
+        while keep_session_alive:
+            res = self._make_authenticated_request(body, headers, url, query, stream)
+            # DSSException 401 is returned and the jwt token expiration is only preserved in the stack trace
+            if not ('jwt.exceptions.ExpiredSignatureError: Signature has expired' in res.json().get('stacktrace', '')):
```
xbrianh: Should we be relying on scraping data from a stack trace? This doesn't feel reliable.

DailyDreaming: I don't like it either, even though I think the edge cases where this is unreliable are unlikely.  I will try to find a better way, but the request itself doesn't currently seem to have any other way to fetch it so I will probably have to rework how the errors are saved and propagated back.

Bento007:
Since OIDC is a standard I wonder if there is a library that will take a JWT with a refresh token and keep it refreshed for you.

hca/util/__init__.py
Bento007: you should inspect the JWT locally to see when and if it's expired.

DailyDreaming:
@Bento007 I believe that's what's already done for sessions instantiated via a Google direct login.  This new method applies only to Google credentials set with the `GOOGLE_APPLICATION_CREDENTIALS` env var and it simply checks the time on the token and refreshes.

hca/util/__init__.py
```diff
@@ -447,14 +446,24 @@ def _get_jwt_from_service_account_credentials(self):
                                 algorithm='RS256').decode()
         return signed_jwt, exp
 
+    def expired_token(self):
```
xbrianh: When the token doesn't exist this method returns `False`, indicating a non-expired token. This may be confusing behavior. It may be preferable to raise in this scenario.

hca/util/__init__.py
```diff
@@ -447,14 +446,24 @@ def _get_jwt_from_service_account_credentials(self):
...
+        """Return True if we have an active session containing an expired (or nearly expired) token."""
+        ten_second_buffer = 10
+        if self._authenticated_session:
+            token_expiration = self._authenticated_session.token.get('expires_at', None)
+            if token_expiration:
+                if token_expiration <= time.time() + ten_second_buffer:
```
xbrianh: What is the rationale for the time buffer versus just letting the token expire? And why 10 seconds?

DailyDreaming: I though it should have _some_ buffer, since I can imagine checking a token and finding it valid with 0.001 seconds left, but when you go to actually make the call it errors because it's expired in that 0.001 seconds.  10 seconds is overkill, but the default expiration is an hour so I figured refreshing 10 seconds early is a comparably negligible buffer with no loss of efficiency that should avoid the situation I described.

hca/util/__init__.py
DailyDreaming: If `self._authenticated_session` doesn't exist, that's fine and it doesn't have an expired token, it has no session (and will create a session outside of this function).  If `self._authenticated_session.token` doesn't exist or is `None`, it will error.

I can enforce that `expires_at` always exists in the token, which it always should and seems like a sensible check.

xbrianh: I believe `self._authenticated_session.token` is initialized to `dict()`, in which case this method will confusingly return `False` if it is used outside of the context of  _your_ use case, which a future coder may plausibly do. Expecting `expires_at` to exist in the token would fix this issue.

DailyDreaming: Changes made.

DailyDreaming: It inits to `None` instead of `dict()`: https://github.com/HumanCellAtlas/dcp-cli/blob/3789edffb6804f02290a905dab093dde6cd7884b/hca/util/__init__.py#L233

test/unit/util/test_swagger_client.py
```diff
@@ -17,6 +17,18 @@
 from unittest.mock import mock_open
 
 
+import time
+from hca.dss import DSSClient
+
+
+class MockDSSClient(DSSClient):
```
chmreid: I don't think this should be called a "mock" DSS client, since it isn't actually mocking anything, just fixing an attribute value. It still works exactly like the normal DSSClient and makes live API calls.

chmreid: maybe rename to `ShortLivedDSSClient` or similar

hca/util/__init__.py
chmreid: Related to Brian's comment - I find it useful to write access control/security functions such that the default boolean returned is "False"/the default outcome is to deny access, and an explicit conditional check must be met for the function to return True and allow access. For example, this function could be renamed to `token_is_active()` or similar, and would return False by default (equivalent to assuming every token is expired) and True if the required conditions are met.

Bento007:
@lon have you seen this object in dcplib https://github.com/HumanCellAtlas/dcplib/blob/5704ea9bc8c07f6ac7905cb63f45e34f3fd4213d/dcplib/security/dcp_service_account.py#L10

test/unit/util/test_swagger_client.py
```diff
@@ -116,6 +128,55 @@ def test_get_swagger_spec_local_config(self):
...
+        r = dss.get_collections()
+        assert 'collections' in r
+        token_two = dss._authenticated_session.token['access_token']
+        expires_at = dss._authenticated_session.token['expires_at'] - time.time()
+        assert expires_at < 1
+
+        assert token_one != token_two  # make sure it requested with two different tokens
+
+    def test_swagger_client_no_refresh(self):
+        """
```
Bento007: I greatly appreciate the test descriptions.

hca/util/__init__.py
Bento007: The security check here is more of a courtesy. Expired tokens cannot be 100% enforced on the client side. I recommend going for readability.

DailyDreaming:
@Bento007 That looks like a nice implementation.  I wish I'd known about it prior.  I think I'll merge since this works, and I'll make a ticket if anyone wants to replace this with the dcplib code in the future.

PR #447, opened 2019-09-26
Adding header values into request when values are present in req_args

Bento007:
Given an OpenAPI specification:
```
...
path:
  /foo:
    get:
      parameters:
        - in: header
          name: bar
          schema:
            type: string
          required: true
...
```
`SwaggerClient.foo(bar="abcd")`, a functions generated after parsing this specification will not apply `bar="abcd" to the header of the request. The following PR fixes this.

### DIFF
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -138,7 +138,7 @@ class _ClientMethodFactory(object):
 
         # TODO: (akislyuk) if using service account credentials, use manual refresh here
         json_input = body if self.body_props else None
-        headers = headers if headers else {}
+        headers = headers or {}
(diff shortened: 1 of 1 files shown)


hca/util/__init__.py
```diff
@@ -139,6 +139,7 @@ def _request(self, req_args, url=None, stream=False, headers=None):
         # TODO: (akislyuk) if using service account credentials, use manual refresh here
         json_input = body if self.body_props else None
         headers = headers if headers else {}
```
xbrianh: Not in the diff, but this could be:
```
headers = headers or {}
```

Bento007: I'll add that in

PR #448, opened 2019-09-26
Misc fixes

Bento007:
- Check for content type application/json before trying to parse.
- Storying the modified `http_path` in `_http_path` to avoid corruption of path url when making requests. This happens in the case where a valid path: `/.well-known/openid-configuration` is replace with the invalid path: `/well_known/openid_configuration` during the request.

### DIFF
--- a/test/integration/util/test_swagger_client.py
+++ b/test/integration/util/test_swagger_client.py
@@ -48,7 +48,10 @@ class TestSwaggerClient(unittest.TestCase):
         "head_with_optional_param",
         "put_with_invalid_enum_param",
         "get_with_allOf_in_body_param",
-        "test_get_with_allOf_multiple_in_body_param"
+        "get_with_allOf_multiple_in_body_param",
+        "get_with_special_characters__in_path",
+        "get_with_header_parameter",
+        "get_with_request_body_application_json"
     ]
 
     @classmethod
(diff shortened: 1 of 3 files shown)


xbrianh:
Would [urllib.parse.urlparse](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) be appropriate for parsing out URL components?

chmreid:
Is it possible to write a test for this? Perhaps in `test/integration/util/test_swagger_client.py`?

Bento007:
@xbrianh the urls is already parsed in the OpenAPI spec.

@chmreid I will add some

test/integration/util/test_swagger_client.py
```diff
@@ -401,6 +405,57 @@ def test_put_with_invalid_enum_param(self):
...
+                                                 url,
+                                                 json=json,
+                                                 stream=stream,
+                                                 params=params,
+                                                 headers=headers,
+                                                 timeout=timeout)
+
+    @staticmethod
+    def get_command(http_method, path):
+        return f"{http_method}{path}".replace('/', '-').replace('_', '-').replace('.', '')
```
chmreid: The operation that cleans the http path above is `http_path.replace('.', '').replace('-', '_')` but that's different from the one here. I think these two should be reconciled by putting this operation in a function in `hca/util/__init__.py` that can also be imported for the tests.

DailyDreaming: Second this.

PR #454, opened 2019-10-03
Prettify the contents of bundle.json

GPelayo:


hannes-ucsc:
@jessebrennan, merge at will.

Issue #427, opened 2019-08-13
Reached link limit when downloading a large amount of bundles (140000+)


GPelayo:
```
WARNING:hca:Download task failed: DSSFile(name='project_0.json', uuid='e0009214-c0a0-4a7b-96e2-d6a83e966ce0', version='2019-07-09T221320.395000Z', sha256='ba4df5b43e0bdff717f6d81b5aaaa941987fb50d3a91498b47a49353abbbedee', size=6366, indexed=True, replica='aws')
Traceback (most recent call last):
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/ubuntu/dcp-venv/lib/python3.6/site-packages/hca/dss/__init__.py", line 580, in _download_and_link_to_filestore
    hardlink(file_store_path, file_path)
  File "/home/ubuntu/dcp-venv/lib/python3.6/site-packages/hca/dss/util/__init__.py", line 57, in hardlink
    os.link(source, link_name)
OSError: [Errno 31] Too many links: '.hca/v2/files_2_4/ba/4df5/ba4df5b43e0bdff717f6d81b5aaaa941987fb50d3a91498b47a49353abbbedee' -> '8ac31d30-3e22-494e-a8c2-0bb837fa7d4e.2019-08-01T200147.756920Z/project_0.json'
```

GPelayo:
I did a quick fix at... https://github.com/HumanCellAtlas/dcp-cli/blob/8322457451d78e17af411a826af462b0647b9e1e/hca/dss/util/__init__.py#L56-L60

...where if an EMLINK exception is thrown, the source file is copied to the link's location instead being linked. This is implemented using the snippet below.

```python
try: 
    os.link(source, link_name)
except OS Error as e:
    if e.errno == errno.EMLINK:
        shutil.copyfile(source, link_name)
    elif e.errno != errno.EEXIST:
        raise
```
While this fix allowed me a large number download files, it was only tested in Ubuntu with Python 3.7 and doesn't completely resolve this issue.

PR #455, opened 2019-10-03
Work around link limit with manifest download (#427)

jessebrennan:


test/unit/test_dss_client.py
```diff
@@ -335,6 +337,20 @@ def _assert_bundle_json(self, prefix=''):
             actual_hash = hashlib.sha256(f.read()).hexdigest()
         self.assertEqual(actual_hash, expected_hash)
 
+    @unittest.skipIf(os.name is 'nt', 'os.link is not used on Windows')
+    def test_link_fails(self):
```
GPelayo: Is it possible to refactor parts of `TestManifestDownloadBundle.test_link_fails()` and `TestDownload.test_link_fails()` into a fixture?

test/unit/test_dss_client.py
```diff
@@ -335,6 +337,20 @@ def _assert_bundle_json(self, prefix=''):
...
+        os_error = OSError()
+        os_error.errno = errno.EMLINK
+        with patch('os.link', side_effect=os_error):
+            from hca.dss.util import log as log_
+            with self.assertLogs(logger=log_, level=logging.WARNING) as logs:
+                self._mock_download_manifest(self.manifest_file, 'aws', layout='bundle')
+        for record in logs.records:
+            self.assertEqual(record.funcName, 'hardlink')
+            self.assertEqual(record.msg, 'Too many hard links to a file, reverting to copying')
+        self.assertEqual(len(logs.records), 9)
```
GPelayo: Indent this line so it's under the ContextManager at line 348 ending in `as logs:`.

jessebrennan: I don't want it under the context manager, because then if that assert statement logs anything, it will change the number of records. Only things that I want to capture the number of records for should be under the context manager.

test/unit/test_dss_client.py
```diff
@@ -125,6 +125,19 @@ def _files_present(self):
                 for dir_path, _, files in walk('.')
                 for f in files}
 
+    def _mock_download_manifest(self, *args, **kwargs):
+        with patch('hca.dss.DownloadContext._download_file', side_effect=_fake_download_file) as download_func, \
+                patch('hca.dss.DSSClient.get_bundle') as mock_get_bundle:
```
hannes-ucsc: I'd break this up into two nested `with` statements. That avoids the weird indention and the backslash.

test/unit/test_dss_client.py
```diff
@@ -125,6 +125,19 @@ def _files_present(self):
...
+            mock_get_bundle.paginate = _fake_get_bundle_paginate
+            self.dss.download_manifest(*args, **kwargs)
+        return download_func
+
+    def _mock_download(self, *args, **kwargs):
+        with patch('hca.dss.DownloadContext._download_file', side_effect=_fake_download_file), \
```
hannes-ucsc: Same here

test/unit/test_dss_client.py
hannes-ucsc: Don't understand what you mean there.

test/unit/test_dss_client.py
hannes-ucsc: It makes sense to minimize the bodies of `with` statements. What would moving the assert into the logs CM buy us?

hannes-ucsc: I'd document why 9 and what it depends on.

test/unit/test_dss_client.py
```diff
@@ -480,6 +480,20 @@ def test_download_dir(self):
...
+        os_error = OSError()
+        os_error.errno = errno.EMLINK
+        with patch('os.link', side_effect=os_error):
+            from hca.dss.util import log as log_
+            with self.assertLogs(logger=log_, level=logging.WARNING) as logs:
+                self._mock_download('any_bundle_uuid', 'aws')
+        for record in logs.records:
+            self.assertEqual(record.funcName, 'hardlink')
+            self.assertEqual(record.msg, 'Too many hard links to a file, reverting to copying')
+        self.assertEqual(len(logs.records), 5)
```
hannes-ucsc: See above.

hca/dss/util/__init__.py
```diff
@@ -51,3 +56,10 @@ def hardlink(source, link_name):
         # Check device first because different drives can have the same inode number
         if source_stat.st_dev != dest_stat.st_dev or source_stat.st_ino != dest_stat.st_ino:
             raise
+    except OSError as e:
+        if e.errno == errno.EMLINK:
+            # FIXME: Copying is not space efficient; see https://github.com/HumanCellAtlas/dcp-cli/issues/453
+            log.warning('Too many hard links to a file, reverting to copying')
```
hannes-ucsc: I'd log the file names unless that's obvious from a previous log statement. But given that this is run in multiple threads I doubt the user can easily correlate log lines from a single thread.

jessebrennan:
@hannes-ucsc I've responded to your review comments, and gone a bit further...

I changed the behaviour so that any errors encountered with linking are logged and then copying is done instead. Let me know if you think this isn't a good idea.

hca/dss/util/__init__.py
```diff
@@ -51,3 +51,7 @@ def hardlink(source, link_name):
         # Check device first because different drives can have the same inode number
         if source_stat.st_dev != dest_stat.st_dev or source_stat.st_ino != dest_stat.st_ino:
             raise
+    except Exception as e:
```
hannes-ucsc: Let's discuss. I'm worried that a systematic error causes all files to be copied.

hannes-ucsc: Can you please reach out to me about this?

hannes-ucsc: @jessebrennan and I talked and agreed to only fall back to aking a copy on EMLINK and to fail hard on all other errors.

PR #457, opened 2019-10-07
PaginatingClientMethodFactory: Use configurable paginated content keys

### LINKED ISSUE #456: DSS GET /bundles/all does not iterate correctly

kislyuk:
Implements support for https://github.com/HumanCellAtlas/data-store/pull/2332.

Fixes #456.

### DIFF
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -189,7 +189,10 @@ class _PaginatingClientMethodFactory(_ClientMethodFactory):
         """
         for page in self._get_raw_pages(**kwargs):
             content_key = page.headers.get("X-OpenAPI-Paginated-Content-Key", "results")
-            for result in page.json()[content_key]:
+            results = page.json()
+            for key in content_key.split("."):
+                results = results[key]
+            for result in results:
                 yield result
 
     def paginate(self, **kwargs):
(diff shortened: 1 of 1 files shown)


hca/util/__init__.py
```diff
@@ -188,15 +188,12 @@ def iterate(self, **kwargs):
...
-            if page.json().get('results'):
-                for result in page.json()['results']:
-                    yield result
-            elif page.json().get('bundle'):
-                for file in page.json()['bundle']['files']:
-                    yield file
-            else:
-                for collection in page.json().get('collections', []):
-                    yield collection
+            content_key = page.headers.get("X-OpenAPI-Paginated-Content-Key", "results")
```
DailyDreaming: I might use something like this until https://github.com/HumanCellAtlas/data-store/pull/2497 percolates up through all of the stages:
```
        for page in self._get_raw_pages(**kwargs):
            results = page.json()
            content_key = page.headers.get("X-OpenAPI-Paginated-Content-Key")
            if content_key:
                for key in content_key.split("."):
                    results = results[key]
                for result in results:
                    yield result
            elif results.get('results'):
                for result in results['results']:
                    yield result
            elif results.get('bundle'):
                for file in results['bundle']['files']:
                    yield file
            else:
                for collection in results.get('collections', []):
                    yield collection
```

Issue #449, opened 2019-09-27
_.csv.Error when using hca dss download-manifest


mshadbolt:
When I try to use hca dss download-manifest the downloads complete but I get an error when it tries to overwrite the original manifest:

```
Traceback (most recent call last):
  File "/Users/mshadbolt/Desktop/beta_user_testing/beta_venv/bin/hca", line 10, in <module>
    sys.exit(main())
  File "/Users/mshadbolt/Desktop/beta_user_testing/beta_venv/lib/python3.7/site-packages/hca/cli.py", line 138, in main
    result = parsed_args.entry_point(parsed_args)
  File "/Users/mshadbolt/Desktop/beta_user_testing/beta_venv/lib/python3.7/site-packages/hca/util/__init__.py", line 587, in arg_forwarder
    return command(**command_args)
  File "/Users/mshadbolt/Desktop/beta_user_testing/beta_venv/lib/python3.7/site-packages/hca/dss/__init__.py", line 334, in download_manifest
    download_dir)
  File "/Users/mshadbolt/Desktop/beta_user_testing/beta_venv/lib/python3.7/site-packages/hca/dss/__init__.py", line 410, in _download_manifest_bundle
    self._write_output_manifest(manifest, download_dir)
  File "/Users/mshadbolt/Desktop/beta_user_testing/beta_venv/lib/python3.7/site-packages/hca/dss/__init__.py", line 693, in _write_output_manifest
    writer.writerow(row)
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/csv.py", line 155, in writerow
    return self.writer.writerow(self._dict_to_list(rowdict))
_csv.Error: need to escape, but no escapechar set
```

mshadbolt:
Just wanted to point out that when I encountered this error I had set up a clean virtual environment and installed hca cli using pip.

I generated the manifest that I used using the prod Data Browser

mshadbolt:
This is the original manifest I downloaded from the browser. I have since tried modifying it and trying to download 1 or 2 files and I always get the same error. I get the same error regardless of if I set the `--layout` to `bundle` or to `none`

[hca-manifest-e101f5e8-e9a2-4ff5-9f5a-62bf14014877.txt](https://github.com/HumanCellAtlas/dcp-cli/files/3663524/hca-manifest-e101f5e8-e9a2-4ff5-9f5a-62bf14014877.txt)

[...]

PR #458, opened 2019-10-09
 Use proper configurations for csv reading/writing (#449) 

jessebrennan:


jessebrennan:
@GPelayo good point!

hca/util/tsv.py
```diff
@@ -0,0 +1,8 @@
+import csv
+import functools
+
+
+# Wrap the csv library with our required options
+
+DictReader = functools.partial(csv.DictReader, delimiter='\t', dialect='excel-tab')
```
hannes-ucsc: Nifty, but I think for type checking, auto-completion and ensuring that callers don't accidentally pass `delimiter` and `dialect` (only to have those be silently ignored) it might be better to make this an explicit def whose signature has only the file name parameter. It's easier to start strict and relax the signature over time than the other way around. I'd be happy to have this be four lines instead of two if I get clarify

jessebrennan:
@hannes-ucsc conflicts are resolved. I'm assigning back to you for re-review, then I can autosquash / merge.

PR #459, opened 2019-10-15
update required python version to 3.5+

mshadbolt:


mshadbolt:
I have no idea if I have done the right thing with requesting this documentation page so if I have done it wrong please let me know the process

PR #465, opened 2019-10-18
auto pagination

### LINKED ISSUE #461: Inline Support for Pagination
As a user if i call the command, `hca dss post-search --replica aws --es-query '{ "query": { "bool": { "must": [ { "match": { "files.project_json.provenance.document_id": "abe1a013-af7a-45ed-8c26-f3793c24a1f4" } } ] } } }' | jq -r '.results[].bundle_fqid' | sed "s/^/\'/" | sed "s/$/\',/" | wc -l` the command will return the default --per-page value of 100, because pagination is supported at this endpoint, all values across all the pages should be concatenated and returned to the user. (in this case 423 values)

The following modification was suggested.
1. update SwaggerClient.build_argparse_subparsers to check if the method supports iteration/pagination
2. if so, pass an “paginate=True” flag to SwaggerClient._command_arg_forwarder_factory
3. in _command_arg_forwarder_factory, in the arg_forwarder closure, if paginate is True, then call command.paginate() in a loop instead of command(); get the paginated content key from the response; and concatenate subsequent pages to that key’s value in the first page response

amarjandu:
Allows for auto pagination for hca commands, turns on this feature by default.
Adds support for optional `--no-paginate` to fall back to single page response. 
Closes #461

### DIFF
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -92,6 +92,7 @@ import argparse
 import time
 import jwt
 import requests
+import jmespath
 
 from inspect import signature, Parameter
 from requests.adapters import HTTPAdapter, DEFAULT_POOLSIZE
@@ -209,8 +210,13 @@ class _PaginatingClientMethodFactory(_ClientMethodFactory):
             if response_data is None:
                 response_data = page
             else:
+
                 content_key = response_data['_internal_api_content_key']
-                response_data[content_key] += page[content_key]
+                data_aggregrator = jmespath.search(content_key, response_data)
+                patch = jmespath.search(content_key, page)
+                if patch:
+                    data_aggregrator += patch
+                response_data[content_key] = data_aggregrator
         response_data.pop('_internal_api_content_key')
         return response_data
 
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -198,7 +198,21 @@ class _PaginatingClientMethodFactory(_ClientMethodFactory):
     def paginate(self, **kwargs):
         """Yield paginated responses one response body at a time."""
         for page in self._get_raw_pages(**kwargs):
-            yield page.json()
+            data = page.json()
+            content_key = page.headers.get("X-OpenAPI-Paginated-Content-Key", "results")
+            data['_internal_api_content_key'] = content_key
+            yield data
+
+    def _cli_call(self, cli_args):
+        response_data = None
+        for page in self.paginate(**vars(cli_args)):
+            if response_data is None:
+                response_data = page
+            else:
+                content_key = response_data['_internal_api_content_key']
+                response_data[content_key] += page[content_key]
+        response_data.pop('_internal_api_content_key')
+        return response_data
 
 
 class SwaggerClient(object):
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -542,10 +542,10 @@ class SwaggerClient(object):
         path_parameters = [p_name for p_name, p_data in parameters.items() if p_data["in"] == "path"]
         self.http_paths[method_name][frozenset(path_parameters)] = http_path
 
+        body_props, method_args = self._process_method_args(parameters=parameters, body_json_schema=body_json_schema)
+
         method_supports_pagination = True if str(requests.codes.partial) in method_data["responses"] else False
         highlight_streaming_support = True if str(requests.codes.found) in method_data["responses"] else False
-
-        body_props, method_args = self._process_method_args(parameters=parameters, body_json_schema=body_json_schema)
 
         factory = _PaginatingClientMethodFactory if method_supports_pagination else _ClientMethodFactory
         client_method = factory(self, parameters, path_parameters, http_method, method_name, method_data, body_props)
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -205,20 +205,24 @@ class _PaginatingClientMethodFactory(_ClientMethodFactory):
             yield data
 
     def _cli_call(self, cli_args):
-        response_data = None
-        for page in self.paginate(**vars(cli_args)):
-            if response_data is None:
-                response_data = page
-            else:
+        if str.lower(vars(cli_args)['page']) == 'false':
+            print('super hit')
+            return super()._cli_call(cli_args)
+        else:
+            response_data = None
+            for page in self.paginate(**vars(cli_args)):
+                if response_data is None:
+                    response_data = page
+                else:
 
-                content_key = response_data['_internal_api_content_key']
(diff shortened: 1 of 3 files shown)


hca/util/__init__.py
```diff
@@ -198,7 +199,29 @@ def iterate(self, **kwargs):
     def paginate(self, **kwargs):
         """Yield paginated responses one response body at a time."""
         for page in self._get_raw_pages(**kwargs):
-            yield page.json()
+            data = page.json()
+            content_key = page.headers.get("X-OpenAPI-Paginated-Content-Key", "results")
+            data['_internal_api_content_key'] = content_key
```
kislyuk: I would prefer it if we could find a way not to pass this in-band.

hca/util/__init__.py
```diff
@@ -198,7 +199,29 @@ def iterate(self, **kwargs):
...
+            yield data
+
+    def _cli_call(self, cli_args):
+        if str.lower(vars(cli_args)['page']) == 'false':
```
kislyuk: Why not just use `cli_args.page` here?

hca/util/__init__.py
```diff
@@ -514,11 +537,14 @@ def _build_client_method(self, http_method, http_path, method_data):
...
         self.http_paths[method_name][frozenset(path_parameters)] = http_path
 
-        body_props, method_args = self._process_method_args(parameters=parameters, body_json_schema=body_json_schema)
-
         method_supports_pagination = True if str(requests.codes.partial) in method_data["responses"] else False
         highlight_streaming_support = True if str(requests.codes.found) in method_data["responses"] else False
 
+        if method_supports_pagination:
+            parameters['page'] = dict(description='Use the internal CLI paging mechanism', required=False,
+                                      name='page', default="True", type=str)
```
kislyuk: Consider using this instead:

add_argument("--no-paginate", action="store_false", dest="paginate")

This will allow you to say `hca dss post-search --no-paginate ...` and the argument to consume will be `args.paginate`, a boolean.

hca/util/__init__.py
amarjandu: I think an alternative way to perform this would be to populate the header value in the `swagger.json` page, then create a internal function that can extract the value from 
`_PaginatingClientMethodFactory.method_data`. This would require changes to the DSS Swagger, and potentially Fusillade/Query Service Swagger pages as well.

hca/util/__init__.py
amarjandu: :) thanks for catching that.

hca/util/__init__.py
amarjandu: Much cleaner, thanks!

amarjandu:
> This needs a test with either a small page size or a mock. I can help with a mock if needed.

I was having a little bit of trouble with the MOCK functionality. I'm having a hard time figuring out where is an effective place to put the mock.  I don't think a small page size would work as `dss-dev` is cleared out every day, which can cause inconsistent/incomplete testing. 

Please let me know if you think the approach to getting the `X-OpenAPI-Paginated-Content-Key` from the `swagger.json` is a better approach.

hca/util/__init__.py
```diff
@@ -198,7 +199,32 @@ def iterate(self, **kwargs):
     def paginate(self, **kwargs):
         """Yield paginated responses one response body at a time."""
         for page in self._get_raw_pages(**kwargs):
-            yield page.json()
+            data = page.json()
+            content_key = page.headers.get("X-OpenAPI-Paginated-Content-Key", "results")
+            yield data, content_key
```
kislyuk: Please don't change the signature of `paginate()`.

hca/util/__init__.py
kislyuk: You can use `_PaginatingClientMethodFactory. _get_raw_pages()`, the method `paginate()` wraps. `_get_raw_pages()` yields entire responses. You can read the paginated content key from the response headers.

kislyuk:
@amarjandu Here is a starting point for a test case that I would use to unit test this with a mock:
```python
import json, io, unittest, unittest.mock
import requests
from requests.models import Response
import hca


class SessionMock:
    status_code = requests.codes.ok

    def request(self, *args, **kwargs):
        payload = {"results": ["foo", "bar"]}
        res = Response()
        res.status_code = self.status_code
        res.headers = {"foo": "bar"}
        res.raw = io.BytesIO(json.dumps(payload).encode())
        return res


class MockTestCase(unittest.TestCase):
    def test_with_mock(self):
        hca.util.SwaggerClient.get_session = SessionMock
        for page in hca.dss.DSSClient().get_bundles_all.paginate(replica="aws"):
            print(page)


if __name__ == "__main__":
    unittest.main()
```
You can do fancier things with `unittest.mock`, but I started simple here and I think we can do all we need here with this approach.

Regarding the paginated content key: please see my other comment about using `_PaginatingClientMethodFactory._get_raw_pages()` in the caller.

amarjandu:
@kislyuk

 I made some changes to this PR. I removed the auto-pagination features from the python bindings, this forces a developer who is using the bindings to go through the `_PaginatingClientMethodFactory.paginate` method, if they want to flex the paging feature. However the when using the cli-commands auto-pagination will be enabled by default.

This could be seen as 'unexpected' behavior, perhaps we should turn off auto-paging in the cli-commands by default, and change the argument to `--paginate`.

I implemented a version of testing with your suggestion, but it is erroring in testing within allspark/travis, when i test locally on my machine the added tests pass. It seems that it is expecting more attributes within the SessionMock to be present, could we change where we are performing the Mock?

kislyuk:
Actually I think we *should* paginate and fetch all results by default in the CLI. I think that's the least surprising behavior.

The reason you're seeing the error is that the test is trying to call `.get()` on the session mock to fetch the Swagger API definition; it's succeeding locally because a copy of the definition has already been cached.

To fix this, define the `get()` method on the mock:
```
class SessionMock:
    ...
    def get(self, *args, **kwargs):
        kwargs["method"] = "GET"
        return self.request(*args, **kwargs)
```

hca/util/__init__.py
```diff
@@ -200,6 +201,31 @@ def paginate(self, **kwargs):
...
             yield page.json()
 
+    def _cli_call(self, cli_args):
+        if cli_args.paginate is not True:
+            return super()._cli_call(cli_args)
+        return self._auto_page(**vars(cli_args))
+
+    def __call__(self, client, **kwargs):
+        # auto pagination should not be available for python bindings, use paginate attribute instead.
+        return super().__call__(client, **kwargs)
```
kislyuk: This `__call__()` declaration seems to be a no-op. Can you remove it?

amarjandu: Resolved.

PR #471, opened 2019-10-25
remove api version from method_name

amarjandu:
allows for better method_names when using open_api 3.0
Before:
```
get-v1-users        List users
post-v1-user        Create a new user
get-v1-user-{user-id}
                        Retrieve information about a user
put-v1-user-{user-id}
                        Modify user status
```

```
get-users           List users
post-user           Create a new user
get-user            Retrieve information about a user
put-user            Modify user status
```

### DIFF
--- a/hca/util/__init__.py
+++ b/hca/util/__init__.py
@@ -92,6 +92,7 @@ import argparse
 import time
 import jwt
 import requests
+import re
 
 from inspect import signature, Parameter
 from requests.adapters import HTTPAdapter, DEFAULT_POOLSIZE
@@ -495,7 +496,8 @@ class SwaggerClient(object):
     @staticmethod
     def _build_method_name(http_method, http_path):
         method_name = http_path.replace('/.well-known', '').replace('-', '_')
-        method_name_parts = [http_method] + [p for p in method_name.split("/")[1:] if not p.startswith("{")]
+        method_name_parts = [http_method] + [p for p in method_name.split("/")[1:] if not p.startswith("{") and
+                                             not re.match(r"[v]\d", p)]
         method_name = "_".join(method_name_parts)
         if method_name.endswith("s") and (http_method.upper() in {"POST", "PUT"} or http_path.endswith("/{uuid}")):
             method_name = method_name[:-1]


kislyuk:
I don't think this is needed for openapi 3.0.

In openapi 3.0, use a full path including the `/vN` prefix as here: https://github.com/HumanCellAtlas/query-service/blob/master/dcpquery-api.yml#L19

In swagger 2.0, use the `basePath` parameter as here: https://github.com/HumanCellAtlas/data-store/blob/master/dss-api.yml#L109.

amarjandu:
> I don't think this is needed for openapi 3.0.
> 
> In openapi 3.0, use a full path including the `/vN` prefix as here: https://github.com/HumanCellAtlas/query-service/blob/master/dcpquery-api.yml#L19
> 
> In swagger 2.0, use the `basePath` parameter as here: https://github.com/HumanCellAtlas/data-store/blob/master/dss-api.yml#L109.

@kislyuk AH i see, `fusillade` uses `url: https://$API_DOMAIN_NAME` as seen [here](https://github.com/HumanCellAtlas/fusillade/blob/0d10388f809e7ccde0aaef17f3d8a0bc57b434de/fusillade-api.yml#L26)

It then describes /v1/ within the actual endpoint, perhaps the appropriate step to make the `hca auth` bindings easier to use would be to refactor `fusillade-api.yml` to be similar to the `queryservice-api.yml`

hca/util/__init__.py
```diff
@@ -495,7 +496,8 @@ def _parse_properties(properties, schema):
     @staticmethod
     def _build_method_name(http_method, http_path):
         method_name = http_path.replace('/.well-known', '').replace('-', '_')
-        method_name_parts = [http_method] + [p for p in method_name.split("/")[1:] if not p.startswith("{")]
+        method_name_parts = [http_method] + [p for p in method_name.split("/")[1:] if not p.startswith("{") and
+                                             not re.match(r"[v]\d", p)]
```
chmreid:
```suggestion
                                             not re.match(r"[v]\d+", p)]
```

chmreid:
```
>>> print(re.match(r"[v]\d","v25")[0])
v2
>>> print(re.match(r"[v]\d+","v25")[0])
v25
```

amarjandu:
Closing this PR in favor of https://github.com/HumanCellAtlas/fusillade/pull/295

PR #476, opened 2019-10-29
Fix version numbering

### LINKED ISSUE #475: Incorrect version number behavior
The version numbers on the releases page (https://github.com/HumanCellAtlas/dcp-cli/releases) and the version number in `setup.py` do not match the version number in `hca/version.py` or the version number returned by `hca --version`.

To reproduce, check out the 6.5.0 tag:

```
git clone https://github.com/humancellatlas/dcp-cli && git checkout v6.5.0 && virtualenv vp && source vp/bin/activate && python setup.py build install && hca --version
```

Now run `hca --version`, which shows version 6.0.1, which is incorrect:

```
hca 6.0.1 (CPython 3.7.4 Darwin-18.6.0-x86_64-i386-64bit)
```

The `hca --version` command uses the version number in `hca/version.py`, which has diverged from the version number in `setup.py`.

chmreid:
This fixes inconsistent version number behavior as mentioned in #475 - also closes #475.

Before merging this pull request, the version number for the master branch should really be changed from "6.5.0" to something like <s>"6.5.0dev"</s> "6.6.0dev" (or similar) that will indicate it is different from the tagged version 6.5.0.


amarjandu:
can we bump up the patch version of it? vs having `6.5.0dev`

chmreid:
> can we bump up the patch version of it?

Do you mean `6.6.0dev` or `5.6.1`?

jessebrennan:
@DailyDreaming I've made no changes to how the versioning was done... I have done a couple releases, but I just followed the procedure in the DCP Ops Wiki. Your vague and unwarranted suspicion is off-putting and befuddling.

amarjandu:
> @DailyDreaming I've made no changes to how the versioning was done... I have done a couple releases, but I just followed the procedure in the DCP Ops Wiki. I find your vague and unwarranted suspicion is off-putting and befuddling.

Screen Shot 2019-10-31 at 9 52 03 AM

If this fix means that we don't have to keep changing this file back to 0.0.0 that would be perfect. I've committed this file before to a random version number, it always manages to sneak into the commit.

DailyDreaming:
@jessebrennan I think it was accidental.  As Amar mentioned, the version in `master` was kept at `0.0.0` but it gets rewritten and accidentally included in commit changes, so we used to have to be wary and make certain that it was changed back.  I've changed it accidentally too, and because there's no documentation around it, I'm sure you (or someone else) accidentally changed it without realizing it.

I think setting it to the actual version with an `rc` or `dev` modifier makes a lot more sense.

chmreid:
Set version to `6.6.0dev` and updated the version check function to see if version ends with `dev`

Issue #450, opened 2019-09-27
download-manifest downloads fastq files twice for analysed projects?


mshadbolt:
I was playing around with downloading files using the `hca dss download-manifest` command and I discovered that when a dataset has been analysed, the fastq files are listed in both the primary and secondary bundles. This causes there to be two lines in the manifest file for the same file. As you are downloading it lists the file as being downloaded twice to the same location which I thought was a bit odd.

```
...
INFO:hca:Download 'Retina_Wong_scRNA_Sample5_I1.fastq.gz' to './double/.hca/v2/files_2_4/cc/6bf0/cc6bf034743ed5bf781320deeceb119b4584a7854d922b59f05392d6cd5cc737'.
INFO:hca:Download 'Retina_Wong_scRNA_Sample5_I1.fastq.gz' to './double/.hca/v2/files_2_4/cc/6bf0/cc6bf034743ed5bf781320deeceb119b4584a7854d922b59f05392d6cd5cc737'.
```

I made a trivial example where I made a tsv manifest with a single file listed and one with a double and compared the run time as well as the size taken up on disk.

When I ran the download with a single fastq file it took 50s, when I ran with two files that are  the same file it took 82s.

When I checked the size of the .hca folder that is downloaded
``` 
(beta_venv) C02YV4HKLVDL:test_dup_download mshadbolt$ du -sh double/
1.6G	double/
(beta_venv) C02YV4HKLVDL:test_dup_download mshadbolt$ du -sh double/.hca
818M	double/.hca
```
But I don't think this is accurate because it double counts when there are hard links I think.

```
(beta_venv) C02YV4HKLVDL:1bfb3f90-6451-4084-a629-00cbc80c3592.2019-09-13T091553.823908Z mshadbolt$ ls -li
total 1655984
(beta_venv) C02YV4HKLVDL:double mshadbolt$ ls -il double/1bfb3f90-6451-4084-a629-00cbc80c3592.2019-09-13T091553.823908Z/
total 1662872
7108801 -rw-------  1 mshadbolt  EBI\Domain Users  847438091 27 Sep 16:21 Retina_Wong_scRNA_Sample5_I1.fastq.gz

(beta_venv) C02YV4HKLVDL:double mshadbolt$ ls -il double/bdb0032a-180d-4810-9f07-3d7e46f2ea19.2019-09-14T035851.501822Z/
total 1675208
[...]

PR #477, opened 2019-11-01
Fix duplicate download (#450)

jessebrennan:
Fixes race in file download where the file in the filestore could be overwritten, which would leave a dangling link to the file outside the filestore and thus a duplicate file.

### DIFF
--- a/hca/dss/__init__.py
+++ b/hca/dss/__init__.py
@@ -21,7 +21,7 @@ import requests
 from atomicwrites import atomic_write
 from requests.exceptions import ChunkedEncodingError, ConnectionError, ReadTimeout
 
-from hca.dss.util import iter_paths, object_name_builder, hardlink
+from hca.dss.util import iter_paths, object_name_builder, hardlink, atomic_overwrite
 from glob import escape as glob_escape
 from hca.util import tsv
 from ..util import SwaggerClient, DEFAULT_THREAD_COUNT
@@ -528,7 +528,7 @@ class DownloadContext(object):
             logger.info("Skipping download of '%s' because it already exists at '%s'.", dss_file.name, dest_path)
         else:
             self._make_dirs_if_necessary(dest_path)
-            with atomic_write(dest_path, mode="wb", overwrite=True) as fh:
+            with atomic_overwrite(dest_path, mode="wb") as fh:
                 fh.write(manifest_bytes)
         file_path = os.path.join(bundle_dir, dss_file.name)
         self._make_dirs_if_necessary(file_path)
@@ -592,14 +592,14 @@ class DownloadContext(object):
         ranged get doesn't yield the correct header, then we start over.
         """
         self._make_dirs_if_necessary(dest_path)
-        with atomic_write(dest_path, mode="wb", overwrite=True) as fh:
+        with atomic_overwrite(dest_path, mode="wb") as fh:
             if dss_file.size == 0:
                 return
 
             download_hash = self._do_download_file(dss_file, fh)
 
             if download_hash.lower() != dss_file.sha256.lower():
-                # No need to delete what's been written. atomic_write ensures we're cleaned up
+                # No need to delete what's been written. atomic_overwrite ensures we're cleaned up
                 logger.error("%s", "File {}: GET FAILED. Checksum mismatch.".format(dss_file.uuid))
                 raise ValueError("Expected sha256 {} Received sha256 {}".format(
                     dss_file.sha256.lower(), download_hash.lower()))
--- a/hca/dss/util/__init__.py
+++ b/hca/dss/util/__init__.py
@@ -3,6 +3,9 @@ import logging
 import os
 import shutil
 from builtins import FileExistsError
+
+import atomicwrites
+
 from ...util.compat import scandir
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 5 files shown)


hca/dss/util/__init__.py
```diff
@@ -51,3 +55,13 @@ def hardlink(source, link_name):
         # Check device first because different drives can have the same inode number
         if source_stat.st_dev != dest_stat.st_dev or source_stat.st_ino != dest_stat.st_ino:
             raise
+
+
+@contextlib.contextmanager
```
hannes-ucsc: This seems to conflate `FileExistsError`s raised by the body of surrounding `with` statement (which shouldn't be suppressed) with those raised by the CM itself. Looking at the code of atomicwrites it is not clear how we could fix this since it uses `contextlib.contextmanager` not the more direct approach of implementing `__enter__` and `__exit__` methods.

jessebrennan: Wow, nice catch. I think I have a solution...

jessebrennan:
@hannes-ucsc tests were failing due to a flaky test. An issue has been created https://github.com/HumanCellAtlas/dcp-cli/issues/490

hca/dss/__init__.py
```diff
@@ -760,14 +759,14 @@ def _write_output_manifest(self):
         fieldnames, source_manifest = self._parse_manifest(self.manifest)
         if 'file_path' not in fieldnames:
             fieldnames.append('file_path')
-        with atomic_write(output, overwrite=True, newline='') as f:
+        if os.path.isfile(output):
```
hannes-ucsc: `isfile` returns True for symlinks as well. You may want to think through that case. I don't remember what `open(p,'w')` does 

Also, if you fail half-way through writing the manifest, you don't want the file to be half-written. Ideally, there should be no file or only a temporary file (not ideal). That's why `atomicwrite` may still be needed here.

hca/dss/util/__init__.py
```diff
@@ -51,3 +55,22 @@ def hardlink(source, link_name):
         # Check device first because different drives can have the same inode number
         if source_stat.st_dev != dest_stat.st_dev or source_stat.st_ino != dest_stat.st_ino:
             raise
+
+
+class atomic_write:
```
hannes-ucsc:
```suggestion
class atomic_overwrite:
```

???

test/unit/test_dss_client.py
```diff
@@ -82,10 +88,12 @@ def _fake_do_download_file_with_barrier(*args, **kwargs):
     fh.write(b'Here we write some stuff so that the fake download takes some time. '
              b'This helps ensure that multiple threads are writing at once and thus '
              b'allows us to test for race conditions.')
+    time.sleep(random.random())  # sleep for some small random amount of time
```
hannes-ucsc: Comment is obvious. I think the string above explains the intent sufficiently.

test/unit/test_dss_client.py
```diff
@@ -29,45 +31,49 @@ def _fake_download_file(*args, **kwargs):
...
-                'sha256': '8ffe4838ac08672041f73f82e5f8361860627271ec31aa479fbb65f2ccc46d05',
-                'size': 9
-            }
-        ]
-    }
-    # This ensures that each bundle.json is distinct
-    for f in bundle_dict['files']:
-        f['bundle_uuid'] = kwargs['uuid']
-    yield {'bundle': bundle_dict}
+def _fake_paginate_factory(fake_hash=False):
```
hannes-ucsc:
```suggestion
def _make_fake_paginate(fake_hash=False):
```

???

hca/dss/util/__init__.py
jessebrennan: Much better! :joy:

test/unit/test_dss_client.py
jessebrennan: Absolutely agree.

test/unit/test_dss_client.py
jessebrennan: Definitely better.

jessebrennan:
@hannes-ucsc to facilitate resolution of merge conflicts, I squashed previous fixups, however, I made sure to save the latest fixups in response to your last review so that it's easy to tell what I changed.

jessebrennan:
@hannes-ucsc in order to facilitate resolution of merge conflicts during rebasing, I had to squash the PR. I kept the review changes from your latest review as separate fixup! commits. I also added a couple new commits to address some broken tests.

test/unit/util/__init__.py
```diff
@@ -17,5 +23,34 @@ def test_dict_merge(self):
         dict3 = {'a': {'b': {'c': 1, 'd': 1}}, 'c': 2, 'd': 2}
         self.assertEqual(hca.util._merge_dict(dict1, dict2), dict3)
 
+
+class TestLinking(TmpDirTestCase):
+    """TmpDirTestCase will ensure any links / files we create are cleaned up"""
```
hannes-ucsc: I don't like one-liner docstrings because I like the vertical space introduced by the two lines with just triple quotes on them. But there is no clear guide for this repo so I'll leave this to you.

hannes-ucsc:
Merge at will.

test/unit/util/__init__.py
jessebrennan: Since this format of docstring is still PEP 8, I think it is acceptable to leave it as is.

PR #482, opened 2019-11-05
Fix import dependencies.

DailyDreaming:
The CLI can't be installed via normal methods anymore since it's importing dependencies when fetching the version.  This should fix: https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/data-store/-/jobs/54102

### DIFF
--- a/hca/config.py
+++ b/hca/config.py
@@ -2,7 +2,6 @@
 import logging
 import os
 
-import tqdm
 from tweak import Config as _Config
 
 
@@ -37,6 +36,7 @@ class ProgressBarStreamHandler(object):
     """
     @staticmethod
     def write(msg):
+        import tqdm  # imported here to avoid unnecessary dependency breaks
         tqdm.tqdm.write(msg, end='')
 
 


DailyDreaming:
@amarjandu It removes it.  I realized that removing that import won't work and am fixing it now with a different solution.

DailyDreaming:
@amarjandu Changed.

hca/config.py
```diff
@@ -37,6 +36,7 @@ class ProgressBarStreamHandler(object):
     """
     @staticmethod
     def write(msg):
+        import tqdm  # imported here to avoid unnecessary dependency breaks
```
chmreid: tqdm is in the requirements.txt, but it looks like only the requirements-dev.txt is installed by the pipeline (https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/data-store/-/jobs/54102). shouldn't the pipeline do a `pip install -r requirements.txt` before running `python setup.py`, which would resolve this problem?

DailyDreaming:
@amarjandu If you run `setup.py` (like we do in smoketest) before you install dependencies (`setup.py` is intended to be run before dependencies so this is unexpected) it will import `__version__` from `version.py` which also imports `__init__.py` implicitly and pulls in the `tqdm` dependency.

This is why it's always good to have blank `__init__.py` files.  I don't think it's a good idea to change now because this would create breaking changes (users could no call `hca.get_config()`) if we completely blanked `__init__.py`.

So the solution is either putting the `tqdm` import in the function or importing the version in a way that doesn't import `hca.__init__.py` (storing it somewhere else or hard-coding the version in `setup.py` rather than importing `version.py`).  It was hard-coded before, which is why it didn't break.

hca/config.py
DailyDreaming: `setup.py` should not have outside dependencies.

PR #483, opened 2019-11-05
Fix import dependencies

DailyDreaming:
Hard-coding this as per the old behavior.  I'll leave it to Charles if he wants to modify this in the future and attempt to adjust the cascade of dependencies that importing `__version__` requires.

### DIFF
--- a/hca/config.py
+++ b/hca/config.py
@@ -2,7 +2,6 @@
 import logging
 import os
 
-import tqdm
 from tweak import Config as _Config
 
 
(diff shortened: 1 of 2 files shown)


chmreid:
I am not a bot

chmreid:
This should probably have been reverted back to `6.5.0` and not `6.6.0dev`: https://github.com/HumanCellAtlas/dcp-cli/pull/476/files#diff-2eeaed663bd0d25b7e608891384b7298L10

PR #491, opened 2019-11-14
Fix some progress bar tests failing silently

natanlao:
Fixes a bug where some progress bar subtests (testing functionality
in a simulated interactive session) would fail silently and
erroneously report success because child process exit code was
not checked.

Also makes some small improvements to test assertion semantics.


chmreid:
@jessebrennan wanted to check in with you, do you see any reason this PR should not be merged?

PR #499, opened 2019-12-16
Allow user-specified bundle uuid for upload().

DailyDreaming:


hca/dss/__init__.py
```diff
@@ -108,7 +109,7 @@ def upload(self, src_dir, replica, staging_bucket, timeout_seconds=1200, no_prog
         Upload a directory of files from the local filesystem and create a bundle containing the uploaded files.
         This method requires the use of a client-controlled object storage bucket to stage the data for upload.
         """
```
jessebrennan: Consider adding:
If no bundle UUID is specified, a random one is used for the new bundle.

DailyDreaming: That already occurs.

DailyDreaming:
@jessebrennan There shouldn't be any problems.  The version is still live so it'll create an updated bundle each time.

PR #504, opened 2020-01-23
Remove references to query service

MDunitz:
- remove query service references and tests
- set default config to point at dss swagger


hca/default_config.json
```diff
@@ -5,7 +5,7 @@
     "swagger_url": "https://dss.data.humancellatlas.org/v1/swagger.json"
   },
   "DCPQueryClient": {
```
amarjandu: key `DCPQueryClient` can be removed from this file

MDunitz: 👍

PR #506, opened 2020-01-28
Close file handlers after use

GPelayo:


hca/dss/upload_to_cloud.py
```diff
@@ -76,44 +76,45 @@ def upload_to_cloud(file_handles, staging_bucket, replica, from_cloud=False, log
...
-                file_uuid = str(uuid.uuid4())
-                key_name = "{}/{}".format(file_uuid, os.path.basename(fh.raw.name))
-                destination_bucket.upload_fileobj(
-                    fh,
-                    key_name,
-                    Config=tx_cfg,
-                    Callback=lambda x: progress.update(x) if log_progress else None,
-                    ExtraArgs={
-                        'ContentType': _mime_type(fh.raw.name),
+        for file_name in file_paths:
```
hannes-ucsc:
```suggestion
        for file_path in file_paths:
```

test/integration/dss/test_dss_api_file_limit.py
```diff
@@ -0,0 +1,32 @@
...
+        uploaded_files = [object_name_builder(p, bundle_path) for p in uploaded_paths]
+        client = DSSClient(swagger_url="https://dss.dev.data.humancellatlas.org/v1/swagger.json")
+        manifest = client.upload(src_dir=bundle_path,
+                                 replica="aws",
+                                 staging_bucket=self.staging_bucket)
+        manifest_files = manifest['files']
+        self.assertEqual(list(file['name'] for file in manifest_files).sort(), uploaded_files.sort())
+
+    @classmethod
+    def tearDownClass(cls):
```
hannes-ucsc: teardown and setup should be symmetric. Either convert this to a setUp() and tearDown() pair—I don't see why this fixture needs to have class scope—or just use a regular `try:` `finally:` construct inside the test itself. If you do that, consider moving this test method to a test case that already has tests for the upload functionality or move the test case to an existing test module with upload functionality if this needs to be its own test case for some reason (I don't currently see why it would.

test/integration/dss/test_dss_api.py
```diff
@@ -418,5 +419,29 @@ def test_python_login_logout_user_account(self):
...
+class TestDssApiFileLimit(unittest.TestCase):
+    staging_bucket = "org-humancellatlas-dss-cli-test"
+    old_open_file_limits = resource.getrlimit(resource.RLIMIT_NOFILE)
+
+    def test_python_open_file_limit_upload(self):
+        old_open_file_limits = resource.getrlimit(resource.RLIMIT_NOFILE)
+        bundle_path = os.path.join(TEST_DIR, "upload", "many-files")
+        uploaded_paths = [x.path for x in iter_paths(str(bundle_path))]
+        uploaded_files = [object_name_builder(p, bundle_path) for p in uploaded_paths]
+        client = hca.dss.DSSClient(swagger_url="https://dss.data.humancellatlas.org/v1/swagger.json")
```
amarjandu: we should avoid uploading test data into `dss-prod`. please change this to `https://dss.integration.data.humancellatlas.org/v1/swagger.json`
@DailyDreaming

DailyDreaming: `https://dss.data.humancellatlas.org/v1/swagger.json` is the default too so this is a no op.

DailyDreaming: `https://dss.data.humancellatlas.org/v1/swagger.json` is the default too so this is a no op.

hannes-ucsc: Let's discuss on Slack. I'll start a thread there.

PR #510, opened 2020-02-10
Pen testing build infrastructure

### LINKED ISSUE #123: Add cli command to download files specified in manifest file constructed by the data browser

### LINKED ISSUE #456: DSS GET /bundles/all does not iterate correctly

GPelayo:


amarjandu:
re-opened to track bundles created.

PR #511, opened 2020-02-10
pin docutils in requirements.txt

amarjandu:
User states that an error occurs when installing the `hca` tool with [conda](https://bioconda.github.io/recipes/hca/README.html?highlight=hca)

Due to conflicting docutils version pinned in `requirements.txt` and from within botocore, conda is unable to install the `hca` tool correctly.

Issue also observed when trying to perform a `pip install -r requirements.txt` within a clean virtual env.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,7 +18,7 @@ dcplib >= 2.0.2, < 3
 
 # The version range of docutils is pinned by botocore. Pinning it here can cause a version conflict.
 # See https://github.com/boto/botocore/pull/1802, https://github.com/HumanCellAtlas/dcp-cli/issues/418
-# docutils
+docutils < 0.16
 
 google-auth >= 1.3.0, < 2
 google-auth-oauthlib >= 0.4.1, < 2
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -18,7 +18,7 @@ dcplib >= 2.0.2, < 3
 
 # The version range of docutils is pinned by botocore. Pinning it here can cause a version conflict.
 # See https://github.com/boto/botocore/pull/1802, https://github.com/HumanCellAtlas/dcp-cli/issues/418
-docutils
+# docutils
```
amarjandu: @kislyuk I'm unsure if this should be handled by commenting out the pinned requirement or removing ln21 entirely; i would like to keep the comment above, but don't want the un-pinning/removal to go unnoticed by another developer.  Any recommendation?

chmreid: This should probably be `docutils < 0.16` as per https://github.com/boto/botocore/commit/a8bbf60b665efee53f5a5d2cbbeb6e5e9e0838f2

amarjandu:
@chmreid im going to wait for EOD today (2-10-20) before merging this and pushing up a new patch, just in-case Andrey had a recommendation.

PR #521, opened 2020-03-24
Fall back to copying when os.link raises error

### LINKED ISSUE #519: Hard-link creation fails when downloading to NFS mounts
I am using hca dss to download data.
hca dss download-manifest --manifest 1M_Neurons.tsv \
    --replica 'aws' \
    --layout bundle

There are lots of hard link problems.

  File "/home/user001/.local/lib/python3.7/site-packages/hca/dss/__init__.py", line 584, in _download_and_link_to_filestore
    hardlink(file_store_path, file_path)
  File "/home/user001/.local/lib/python3.7/site-packages/hca/dss/util/__init__.py", line 50, in hardlink
    os.link(source, link_name)
PermissionError: [Errno 1] Operation not permitted: '.hca/v2/files_2_4/8f/aec6/8faec66817969ae6f847b0c649e7328af6085d88e71b32b0e3a8284df4cd88f7' -> '33855cf6-6f3e-4b8f-9cb4-c2b2ea9f528d.2019-05-16T211813.099000Z/dissociation_protocol_0.json'

My file systems do not support hard link across directories. Is there an option to use soft link?

Why is the data downloaded to .hca, and then link to the target directory using hardlink? Can hca download directly to the target directory or move the files from .hca to the target directory?

chmreid:
This addresses a permission denied error on NFS file systems that is caused by the `os.link` call. Addresses #519 

Previously we were catching a "too many links" exception, now we are also catching a "permission denied" exception and an "operation not permitted" exception. (Note that we may be able to remove the "operation not permitted" exception number from the list of exceptions that lead us to fall back to copying.)

### Useful Links

Useful discussion of possible approaches to solving this better: https://bugs.python.org/issue37612

Useful info about NFS and Python's os module: https://web.archive.org/web/20100912144722/http://www.unixcoding.org/NFSCoding

Note: this PR does not (yet) have tests for the new error codes being caught/handled. Tests TBA.

### DIFF
--- a/hca/dss/util/__init__.py
+++ b/hca/dss/util/__init__.py
@@ -58,9 +58,10 @@ def hardlink(source, link_name):
         if source_stat.st_dev != dest_stat.st_dev or source_stat.st_ino != dest_stat.st_ino:
             raise
     except OSError as e:
-        # ELINK can occur when max number of links has been reached
-        # EPERM can occur when writing to NFS
-        copy_on_error = (errno.EMLINK, errno.EPERM)
+        copy_on_error = (
+            errno.EMLINK,  # max. number of hard links exceeded
+            errno.EPERM,  # observed on NFS mounts (issue #519)
+        )
         if e.errno in copy_on_error:
             # FIXME: Copying is not space efficient; see https://github.com/HumanCellAtlas/dcp-cli/issues/453
             log.warning('Failed to link source `%s` to destination `%s`; reverting to copying', source, link_name)
--- a/hca/dss/util/__init__.py
+++ b/hca/dss/util/__init__.py
@@ -58,11 +58,9 @@ def hardlink(source, link_name):
         if source_stat.st_dev != dest_stat.st_dev or source_stat.st_ino != dest_stat.st_ino:
             raise
     except OSError as e:
-        copy_on_error = [
-            errno.EMLINK,  # too many links
-            errno.EACCES,  # permission denied
-            errno.EPERM,  # operation not permitted
-        ]
+        # ELINK can occur when max number of links has been reached
+        # EPERM can occur when writing to NFS
+        copy_on_error = (errno.EMLINK, errno.EPERM)
         if e.errno in copy_on_error:
             # FIXME: Copying is not space efficient; see https://github.com/HumanCellAtlas/dcp-cli/issues/453
             log.warning('Failed to link source `%s` to destination `%s`; reverting to copying', source, link_name)
--- a/hca/dss/util/__init__.py
+++ b/hca/dss/util/__init__.py
@@ -58,7 +58,13 @@ def hardlink(source, link_name):
         if source_stat.st_dev != dest_stat.st_dev or source_stat.st_ino != dest_stat.st_ino:
             raise
     except OSError as e:
-        if e.errno == errno.EMLINK:
+        # See https://docs.python.org/3/library/errno.html
+        copy_on_error = [
+            errno.EMLINK,  # too many links
+            errno.EACCES,  # permission denied
+            errno.EPERM,  # operation not permitted
+        ]
+        if e.errno in copy_on_error:
             # FIXME: Copying is not space efficient; see https://github.com/HumanCellAtlas/dcp-cli/issues/453
             log.warning('Failed to link source `%s` to destination `%s`; reverting to copying', source, link_name)
             shutil.copyfile(source, link_name)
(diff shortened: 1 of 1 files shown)


jessebrennan:
@chmreid I see that you requested my review, but didn't assign the PR to me. I'm confused by this. Did you see this Slack thread? https://ucsc-cgl.slack.com/archives/C705Y6G9Z/p1585093339100200

chmreid:
@jessebrennan The problem is with the `master` branch, nothing specific to this PR. I tried re-running Travis tests for master and they failed: https://travis-ci.org/github/HumanCellAtlas/dcp-cli/builds/654957449

We can pick up on this PR once master is passing.

chmreid:
@jessebrennan Tests are passing, can you give this a review now?

hannes-ucsc:
I would like the opportunity to review this. I'm currently not assigned so I'll hold off for now.

chmreid:
I have assigned this issue to you, please review when you get the chance. Thanks!

hca/dss/util/__init__.py
```diff
@@ -58,7 +58,13 @@ def hardlink(source, link_name):
         if source_stat.st_dev != dest_stat.st_dev or source_stat.st_ino != dest_stat.st_ino:
             raise
     except OSError as e:
-        if e.errno == errno.EMLINK:
+        # See https://docs.python.org/3/library/errno.html
```
hannes-ucsc, line 61:
```suggestion
```

Redundant, any Python programmer can locate the documentation for the standard library. We don't link to `logging`, `shutil` etc. and `errno` is no different.

hca/dss/util/__init__.py
```diff
@@ -58,7 +58,13 @@ def hardlink(source, link_name):
...
+        copy_on_error = [
+            errno.EMLINK,  # too many links
+            errno.EACCES,  # permission denied
+            errno.EPERM,  # operation not permitted
```
hannes-ucsc, line 65:
```suggestion
            errno.EMLINK,
            errno.EACCES,
            errno.EPERM,
```

The meaning of the constants is [well documented](https://docs.python.org/3.6/library/errno.html#errno.EMLINK). We don't need to duplicate that here. What isn't obvious is the intent e.g., *why* EPERM is included. Like that we expect (and have observed) this to occur on an NFS share.

hca/dss/util/__init__.py
```diff
@@ -58,7 +58,13 @@ def hardlink(source, link_name):
         if source_stat.st_dev != dest_stat.st_dev or source_stat.st_ino != dest_stat.st_ino:
             raise
     except OSError as e:
-        if e.errno == errno.EMLINK:
+        # See https://docs.python.org/3/library/errno.html
+        copy_on_error = [
```
hannes-ucsc, line 62: We typically use tuples for immutable sequences that are used with `in`. Not for performance reasons but to more clearly indicate intent.

hca/dss/util/__init__.py
```diff
@@ -58,7 +58,13 @@ def hardlink(source, link_name):
...
+            errno.EMLINK,  # too many links
+            errno.EACCES,  # permission denied
```
hannes-ucsc, line 64:
```suggestion
```

Why is EACCES included here? Our current philosophy is to catch only expected errors that we know can be addressed by falling back to a copy. I have seen no evidence that EACCES is to be expected and should be handled by falling back to a copy. Mere similarity isn't sufficient evidence for me. In fact, I suspect that it is very likely that if `os.link` fails with EACCES, so will `open(…, 'w')` which is what `shutil.copyfile` uses under the hood.

hca/dss/util/__init__.py
chmreid: Comment removed

hca/dss/util/__init__.py
chmreid: Comments removed, better descriptive (why EPERM or EMLINK) added

hca/dss/util/__init__.py
chmreid: Changed to tuple

hca/dss/util/__init__.py
chmreid: This is a good point, removed EACCES so now it is just EMLINK and EPERM

chmreid:
@hannes-ucsc Thank you for the comments. I have addressed your feedback, and this PR is ready for another review. Would you please review it at your convenience? Thanks.

PR #522, opened 2020-05-07
Remove dss-prod from testing

### LINKED ISSUE #123: Add cli command to download files specified in manifest file constructed by the data browser

### LINKED ISSUE #456: DSS GET /bundles/all does not iterate correctly

amarjandu:
Removes the potential for dss-tests to be run against `dss-prod` when on an operators machine. 

See: https://github.com/DataBiosphere/azul/issues/1768

### DIFF
--- a/test/integration/dss/test_dss_cli.py
+++ b/test/integration/dss/test_dss_cli.py
@@ -27,20 +27,12 @@ class TestDssCLI(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         """
-        These tests rely on a logged in env, assert that the stage is not dss-prod
+        These tests rely on a logged in env, ensure that the current stage is not dss-prod
         """
-        # the dss-api endpoint 'get-bundles-all' provides the stage within the response
-        args = ["dss", "get-bundles-all", "--replica", "aws",
-                "--per-page", "10", "--no-paginate"]
-        with CapturingIO('stdout') as stdout:
-            hca.cli.main(args)
-        result = json.loads(stdout.captured())
-        try:
-            assert("https://dss.data.humancellatlas.org" not in result["dss_api"])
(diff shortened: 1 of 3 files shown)


test/integration/dss/test_dss_cli.py
```diff
@@ -24,6 +24,24 @@
...
+                "--per-page", "10", "--no-paginate"]
+        with CapturingIO('stdout') as stdout:
+            hca.cli.main(args)
+        result = json.loads(stdout.captured())
+        try:
+            assert("https://dss.data.humancellatlas.org" not in result["dss_api"])
+        except AssertionError:
+            print('\nPlease change the DSS swagger endpoint in the HCA config file away from `dss-prod`\n'
+                  'For more information see: `https://github.com/HumanCellAtlas/dcp-cli#development`')
+            exit(1)
```
jessebrennan, line 43: Couldn't you also do this by inspecting / importing the DSS config? Then you could check the stage without needing to make an API call.

Also I think this is a round-about use of assert. A conditional would be more concise and clear.

Calling exit here seems like overkill. If only these tests use the outside resource, then there's no need to quit the whole process; other tests should be allowed to run. If other tests (not in TestDssCLI) are counting on this fail mechanism then I'd say that's a problem. Unittest doesn't guarantee the order or test evaluation.

Since you don't need to call exit(), then I think it's sufficient just to `raise ValueError()` here. This would allow other tests (if they exist) to still run.

amarjandu: Fair points, I've incorporated them all. :)

hannes-ucsc:
If the association to the connected issue isn't clear: this PR prevents the creation of more bad bundles by DSS tests. And it's the goal of #1768 to remove those bundles.

PR #523, opened 2020-05-08
create sub-directories when downloading files with !

amarjandu:


test/integration/dss/test_dss_api.py
```diff
@@ -21,6 +21,18 @@
 from test import reset_tweak_changes, TEST_DIR
 
 
+def get_uploaded_file_names(path: str):
+    files = []
+    for entry in os.scandir(path=path):
+        if entry.is_file():
+            files.append(entry.name)
+        if entry.is_dir():
```
hannes-ucsc, line 29:
```suggestion
        elif entry.is_dir():
```

and `else: assert False` at the end.

test/integration/dss/test_dss_api.py
```diff
@@ -21,6 +21,18 @@
...
+            nested_files = get_uploaded_file_names(entry.path)
```
hannes-ucsc, line 30: This is building an excessive number of temporary strings and lists during recursion. Think about how you can make this more efficient. Also check if there isn't built-in functionality for performing a recursive listing of files.

test/integration/dss/test_dss_api.py
```diff
@@ -71,9 +83,31 @@ def test_python_nested_bundle_upload_download(self):
...
+        bundle_uuid = bundle_output['bundle_uuid']
+        bundle_version = bundle_output['version']
+        bundle_fqid = "{}.{}".format(bundle_uuid, bundle_version)
+
+        manifest_files = bundle_output['files']
+        self.assertEqual({file['name'] for file in manifest_files}, uploaded_files)
+
+        with tempfile.TemporaryDirectory() as dest_dir:
+            self.client.download(bundle_uuid=bundle_output['bundle_uuid'], replica="aws", download_dir=dest_dir)
+            nested_downloaded_files = [file.name for file in os.scandir('{}/{}/zarr'.format(dest_dir, bundle_fqid))]
```
hannes-ucsc, line 104: Do you really have to list all files to assert that two of them were downloaded? What about `os.path.isfile`?

